PR #5, opened 2019-06-07
Changed sync stdin to async stdin

tsathishkumar:
This simple fix should allow readying user input while traversing the filesystem. I haven't tested it though!

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -41,7 +41,7 @@ fn run() -> Result<(), Error> {
                     .with_context(|_| "Interactive mode requires a connected terminal")?
             };
             let mut app = TerminalApp::initialize(&mut terminal, walk_options, paths_from(input)?)?;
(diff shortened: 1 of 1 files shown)


Byron-TW:
Thanks a bunch, it's good to know there is some support within Termion - will have a look at the example! Indeed, right now everything is synchronous. However, I am afraid it takes more than that to allow asynchronous operation.
There is a separate function, `initialize()`, which deals with traversal and GUI updates during traversal, and all that even without processing events. Thus this change will have no effect at all.

If you like you can transform this into an issue - indeed, async operation would be nice.

Byron-TW:
I had a look at the example and believe the `async_stdin` capability can truly help - it definitely needs additional implementation and testing, but I am happy to give it a try at some point.

Byron-TW:
Closed in favour of #6 .

PR #9, opened 2019-06-07
More navigation key options

tsathishkumar:
Fixes Left, Right key inputs and Adds Left, Right for Ascent and Descent navigation as it feels like more natural.


Byron:
Thanks a lot for your help! I hope by tonight I finally got a version that can delete things as well.

PR #11, opened 2019-06-15
error formatting suggestions

oconnor663:
We can avoid the `(s)` by inspecting the number. And using spaces instead of tabs makes the output more consistent in a terminal, no longer depending on the length of the preceding filename.


Byron-TW:
Thanks a lot! Much appreciated!

PR #25, opened 2019-07-24
Fix color scheme for light terminals

### LINKED ISSUE #13: _dua interactive_ only looks good with dark-themed terminals
Compare the images

light
dark

vks:
Fixes #13.

Due to the way colors work in the terminal, I had to change the color scheme for dark terminals as well.

I also tried to make the coloring of the entries and the mark pane more consistent.


Byron-TW:
Awesome, thank you so much! I will cut a new release today :).

PR #27, opened 2019-07-25
Fix treatment of symlinks in interactive mode

vks:
Make sure to use `symlink_metadata` instead of `metadata`. This makes sure that symlinks are not followed when recursively deleting directories or when calculating file sizes.

The non-interactive mode is unaffected, because `jwalk` does not provide options to deal with symlinks.


Byron-TW:
Thanks so much! This is a big one, despite the small changes. It puzzles me how I could not find this in the documentation, it's quite obvious actually.

I will cut a new release right away!

PR #35, opened 2020-02-22
Add support for real/apparent size

Freaky:
This adds support for retrieving the real on-disk size of files using the `filesize` crate, leaving the previous behaviour under `-A`/`--apparent-size` flags to match BSD/GNU `du`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -191,6 +191,7 @@ dependencies = [
  "byte-unit",
  "failure",
  "failure-tools",
+ "filesize",
  "itertools",
  "jwalk",
  "log",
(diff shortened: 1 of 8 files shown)


Freaky:
To illustrate the difference:

```
-% du -s --si target
367M    target
-% du --si -ls target
425M    target
-% target/debug/dua -A target
 853.28 MB target
-% target/debug/dua target
 419.09 MB target
```

Not sure where those other few MB are going, but the bulk of the difference now is from `dua` not tracking hardlinks.

Byron:
Thanks so much!

I wouldn't have thought it will be that easy, and that this doesn't affect performance noticeably (at least by the highly unscientific tests I have performed).

By the time you read this, there is already a new release, v2.2, containing this change.

Freaky:
It's free on Unix, part of the metadata structure.  On Windows it'll add another syscall.

Issue #36, opened 2020-02-22
Hardlinked files are not counted properly


Freaky:
The same file can appear in multiple places in a directory hierarchy.  This includes things like Rust's own build directories, `/rescue` partitions with tools like busybox, backups using rsync `--link-dest`, and so forth.

Traditional `du`'s deal with this by keeping a map of seen inodes and their link counts, and avoid counting additional copies of the same inode - decrementing the link count until all have been seen and the inode can be forgotten.  This may not be perfect for an interactive tool like this, but it would probably be better than nothing.

Byron:
Thanks a lot for posting your insights on how this is actually implemented in `du`. This might be helpful when trying to resolve this issue.

Performance-wise it might be OK to implement it, even though the memory consumption might go up visibly. However, being correct AND fast (enough) by default is my preference, too, and as long as hardlink tracking can be disabled I would be happy to see this fixed.

Freaky:
```shell
-% target/debug/dua -l /rescue
   1.14 GB /rescue
-% target/debug/dua /rescue
  14.22 MB /rescue
```

:)

Byron:
**cannot** wait :)))!

Freaky:
Hang on while I beat git into submission and repaint part of the bikeshed :)

Byron:
Closed with release v2.3.0 :)!

PR #37, opened 2020-02-22
Add hardlink tracking, and an option to disable it

Freaky:
This adds a simple `HashMap`-based inode filter, and `-l` and `--count-links` flags to disable it, again mirroring BSD and GNU `du`.

Windows support needs work since it's using unstable features, but the build appears broken there anyway.

### DIFF
--- a/src/inodefilter.rs
+++ b/src/inodefilter.rs
@@ -1,11 +1,10 @@
-
 #![cfg_attr(windows, feature(windows_by_handle))]
 
 use std::collections::HashMap;
 
 #[derive(Debug, Default, Clone)]
 pub struct InodeFilter {
-    inner: HashMap<u64, u64>
+    inner: HashMap<u64, u64>,
 }
 
 impl InodeFilter {
@@ -46,7 +45,7 @@ impl InodeFilter {
                 }
 
                 false
-            },
+            }
             None => {
                 self.inner.insert(inode, nlinks - 1);
                 true
--- a/src/inodefilter.rs
+++ b/src/inodefilter.rs
@@ -1,0 +1,76 @@
+
+#![cfg_attr(windows, feature(windows_by_handle))]
+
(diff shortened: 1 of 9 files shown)


Freaky:
I should probably have mentioned #36 somewhere.

Freaky:
```
-% target/debug/dua target
 309.25 MB target
-% du -s --si target
313M    target
-% du -sh target
299M    target
-% target/debug/dua -f MiB target
  294.93 MiB target
```

Close but not quite.

src/inodefilter.rs
```diff
@@ -0,0 +1,76 @@
+
+#![cfg_attr(windows, feature(windows_by_handle))]
+
+use std::collections::HashMap;
+
+#[derive(Debug, Default, Clone)]
+pub struct InodeFilter {
+    inner: HashMap<u64, u64>
```
Byron: Just out of curiousity: Isn't a BTreeMap faster in this case? The way I understand it: HashMaps first hash their input, and then stuff that into a tree to do comparisons. u64 naturally has that ability, so the Hash part seems superfluous.
What am I missing?

Byron: It seems that a BTreeMap is actually a tad slower, but could be within the margin of error. My dataset took 35s to iterate with default options, and the BTreeMap came in ~400ms slower.

Freaky: I suspected a tree might be slower due to either more pointer chasing or the cost of its linear searches in buckets vs the hashmap's direct indexing, but wasn't sure either way, and it could easily vary across systems depending on the relative performance of CPU/memory.

The HashMap was the more conservative option, since the additional features of a tree weren't needed, and I doubt the difference matters much at this scale.

Byron: Thanks for sharing! As always, much appreciated, I learned something.

PR #40, opened 2020-02-25
Add Fedora installation instructions

tim77:
Happy to say that `dua-cli` packaged finally and [available](https://src.fedoraproject.org/rpms/rust-dua-cli) in official Fedora repos. At this moment only for Rawhide branch but soon will be for current release.


Byron:
Fantastic, thanks so much!

PR #46, opened 2020-03-23
Update to filesize v0.2

Freaky:
No functional changes, but a bit tidier with this interface I think.


Byron:
Thanks a lot for contributIng :)! 
I will see to cutting a new release with updated dependencies.

PR #52, opened 2020-04-03
Disable default features for tui in tui-react

mibac138:
The only default feature in tui is the termion backend, but tui-react doesn't depend on any backend. With the termion backend enabled by default you can't change the backend to another one in other crates.


Byron:
Thanks a lot! I will publish a new minor release of tui-react soon.

Byron:
Done

Issue #58, opened 2020-06-30
Integer overflow with huge files


Freaky:
When faced with file sizes that sum beyond the limits of a u64, such as:

```
.rw-r--r-- 9.2E freaky freaky 29 Jun 21:48 huge1
.rw-r--r-- 9.2E freaky freaky 29 Jun 21:49 huge2
.rw-r--r-- 9.2E freaky freaky 29 Jun 21:49 huge3
.rw-r--r-- 9.2E freaky freaky 29 Jun 21:49 huge4
```

`dua -A` in debug mode panics:

```
thread 'main' panicked at 'attempt to add with overflow', /home/freaky/code/dua-cli-freaky/src/aggregate.rs:60:21
```

And produces incorrect results in release builds:

```
  18.45 EB huge
```

Not really worse than most other similar tools - GNU `du` just gives up and declares the size infinite, and BSD `du` seems to believe negative file sizes are possible - but perhaps it would be nice to Just Work, given Rust's u128 support.

Byron:
You have just shattered my idea that nobody, ever, would hit the limits of u64, who can fit the amount of stars in the known universe multiple times and all :D. Probably that idea of mines is quite similar to that other one, along the lines of "640kb of main memory is going to be enough" :D.

Out of curousity: How did you produce these files? Are they actually backed by that much disk space?

To the point: I think it would not be an issue to represent the final aggregation as u128.
Would you be interested in giving that a shot?

Freaky:
```shell
 ❯ truncate -s 9223372036854775807 huge1
 ❯ l huge1
.rw-r--r-- 9.2E freaky freaky  1 Jul 15:56 huge1
```

They're sparse files and consume no more space than an empty file.

I'll have a look and see if I get bored before I finish fixing the compiler errors ;)

Freaky:
Test environment:

```shell
 ❯ truncate -s 9223372036854775807 huge/{1,2,3,4,5,6,7,8,9}{0,1,2,3,4,5,6,7,8,9}{0,1,2,3,4,5,6,7,8,9}
```

Before:

```shell
 ❯ dua -A huge
  18.45 EB huge
```

After:

```shell
❯ dua -A huge
   9.22 ZB huge
```

Byron:
Wonderful! I wasn't aware that `truncate` could also create files - thus far I was just using it for removing a trailing newlines using `-s -1` 😅.
Amazing.

Closing this issue as the PR fixing it was merged.

PR #59, opened 2020-07-01
Use u128 for byte sizes

Freaky:
Per issue #58, u64 is insufficient for use with very large sparse files.
Enormous file sizes are also a common filesystem error trope, either
from disk corruption or software bugs, and they're also conceivable with
virtual filesystems.

Handle this as gracefully as can be reasonably expected using 128-bit
integers, which should be sufficient for most uses.


Freaky:
Sure :)

Byron:
And it's done :): b46cd3a4920155cffbaecaf1ec8efe0ec245c531

Issue #53, opened 2020-05-20
Interactive mode: subdirectories are not sized in Windows 10


mardukbp:
When executing `dua i` inside a directory in Windows 10 all subdirectories are reported as having a size of 0.0 Bytes.

Byron:
Thanks for reporting. That's an odd one, as directories are computed by looking at the sizes of all files contained within, recursively.
If files have a size, the directories should also have a size no matter which OS `dua` runs on.

Since I cannot try it on windows myself, a fix would have to be contributed.

rivy:
I've found the problem at https://github.com/Byron/dua-cli/blob/master/src/traverse.rs#L108.

`data.name` only contains the file name, not the full path, so `data.name.size_on_disk_fast(m)...` always ends up returning a `0` file size.

I'll post a PR with a fix tomorrow, if you'd like; I need to test the change on *nix machine (my WSL instance doesn't format correctly).

Byron:
It's interesting, as the apparently working counterpart for the non-interactive version [indeed uses the whole path](https://github.com/Byron/dua-cli/blob/82d005b9e3ed9ce8d4441c607ec160f2f0a48b1c/src/aggregate.rs#L45). This might already be the fix here as well.

Ok, I had to try it out - and it really does want to have the filename in the entry, as it will build the path dynamically via the tree - this saves memory.

Sorry for not waiting for the PR, I was just too intrigued to finally see this issue go, but I would be happy if you could give it another round of testing on windows to verify this issue is fixed for good.

Thanks so much for your investigation and all your help!

rivy:
> It's interesting, as the apparently working counterpart for the non-interactive version [indeed uses the whole path](https://github.com/Byron/dua-cli/blob/82d005b9e3ed9ce8d4441c607ec160f2f0a48b1c/src/aggregate.rs#L45). This might already be the fix here as well.

Yep, that was the "working baseline" that I used to trace down the bug and figure out the correction.

> Ok, I had to try it out - and it really does want to have the filename in the entry, as it will build the path dynamically via the tree - this saves memory.
> 
> Sorry for not waiting for the PR, I was just too intrigued to finally see this issue go, but I would be happy if you could give it another round of testing on windows to verify this issue is fixed for good.

No worries. I was fixing it in a different manner... your fix obviously better matches your coding style. 😄 

And, it's working on my windows machine! 🚀

> Thanks so much for your investigation and all your help!

Happy to help.

[...]

PR #60, opened 2020-07-06
Fix windows compiler warnings + add usual wildcard support

rivy:


Byron:
Thanks a million!
I would love to cut a release, but would hold out on it if you see a chance to get #53 fixed. It's probably something small and silly, but inarguably has a huge payout :).

rivy:
I've just had some time to track down the problem for the #53 issue (see #53 for details).

I'll try to post a PR with that fix tomorrow.

PR #61, opened 2020-07-22
Fix disabled tests

rivy:
- fixes 'sample_01_tree' size miscalculation; symbolic links are '0' size on windows vs '1' size on *nix-types
- fixes 'sample_02_tree' path construction error

Both tests now complete successfully on Windows and *nix-type platforms.

### DIFF
--- a/src/interactive/app_test/unit.rs
+++ b/src/interactive/app_test/unit.rs
@@ -17,8 +17,6 @@ fn it_can_handle_ending_traversal_reaching_top_but_skipping_levels() -> Result<(
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


Byron:
Thanks again! You made me see how biased I am towards not investigating these kinds of issues anymore, so I am glad you took the time to bring these tests to windows as something that is very doable, too.

I am currently updating dependencies, and trying to migrate to tui 0.10 is quite some work, probably work worth doing given that TUI is the only library of its kind right now and I don't want to loose access to future improvements.

When done, I will cut a new release.

PR #63, opened 2020-07-27
src, aggregate: fix colors for aggregate mode

Mephistophiles:
Use cyan color for folders in aggregate mode

Before:
before

After:
after


shilangyu:
Much better! 

Before:

image

After:

image

Byron:
Thanks a lot for your contribution, it's much appreciated! Somehow I could never make myself use cyan more (the color scheme works for me), so I am glad low contrast and disappearing text is finally fixed.

Release v2.10.2 is now (should soon) be available.

PR #74, opened 2021-01-07
CHANGELOG: fix typo

ngirard:
Happy new year, and thanks for all the fish !

Cheers


Byron:
Thanks:). Even though I don’t get the fish reference 😅.

ngirard:
See https://en.wikipedia.org/wiki/So_Long,_and_Thanks_for_All_the_Fish ;-)

PR #81, opened 2021-02-20
Respect 'stay_on_filesystem' when no input files are provided

vlakreeh:
Currently with no input files provided dua will traverse all entries in the current directory including different devices, even if the `stay_on_filesystem` flag is set. This PR assumes that with the flag set the user doesn't want to traverse different devices in the current working directory.


Byron:
Thanks a lot for the improvement! It will be available in v2.11.1.

Issue #94, opened 2021-06-23
Enhancement request: move to trash instead of removing


FedericoStra:
Hello, very nice application, I discovered it today and already know that I'll use it for the rest of my life!

Just a little idea: would it be possible to add the option to move marked files to trash instead of permanently removing them? Probably something like https://crates.io/crates/trash might be useful. The command could be bound to `Ctrl+t` in the mark pane for instance.

In the weekend I might have time to try and propose a PR.

**Edit:** a few notes after a brief glance at the code, for my future self and others.

- In function [`AppState::dispatch_to_mark_pane`](https://github.com/Byron/dua-cli/blob/main/src/interactive/app/handlers.rs#L199-L239) add an arm `Some(MarkMode::Trash)` to the [`match`](https://github.com/Byron/dua-cli/blob/main/src/interactive/app/handlers.rs#L211).
- Implement `AppState::trash_entry` in addition to [`AppState::delete_entry`](https://github.com/Byron/dua-cli/blob/main/src/interactive/app/handlers.rs#L211), using https://crates.io/crates/trash.
- Add a `Trash` variant to the `enum` [`MarkMode`](https://github.com/Byron/dua-cli/blob/main/src/interactive/widgets/mark.rs#L29-L31).
- Add an arm `Ctrl('r') => return Some(self.prepare_trashing())` to the `match` in [`MarkPane::process_events`](https://github.com/Byron/dua-cli/blob/main/src/interactive/widgets/mark.rs#L108-L124).
- Implement `MarkPane::prepare_trashing` in addition to [`MarkPane::prepare_deletion`](https://github.com/Byron/dua-cli/blob/main/src/interactive/widgets/mark.rs#L180-L186), which returns the `MarkMode::Trash` variant.

Byron:
> Hello, very nice application, I discovered it today and already know that I'll use it for the rest of my life!

If that isn't a wonderful compliment then I don't know what is :) - thank you, I am glad you like it as much as I do, being an avid user myself.

> Just a little idea: would it be possible to add the option to move marked files to trash instead of permanently removing them? Probably something like https://crates.io/crates/trash might be useful. The command could be bound to Ctrl+t in the mark pane for instance.

[...]

PR #95, opened 2021-06-27
Implement Ctrl+t move to trash

FedericoStra:
Hello,

this is a very quick initial implementation of `Ctrl + t` (move to trash), addressing #94.

There is some code duplication, some minor details are missing (for instance, in the header of the mark pane there is only the reminder for `Ctrl + r` and not for `Ctrl + t`: https://github.com/Byron/dua-cli/blob/main/src/interactive/widgets/mark.rs#L333-L375), and there are no tests of this new feature.

I have ideas to reduce the duplication, but I didn't want to be too invasive with my modifications.
Let me know what I have to change/refine before this PR can be merged.


Byron (changes requested):
Thanks a lot! I think it's a great start and proof that it can work.

The refactoring needed to deduplicate most of the code would definitely be appreciated, and adding `ctrl+t` to the text shouldn't be a problem either.

Once these changes are done it should be ready for merge.

src/interactive/app/handlers.rs
```diff
@@ -247,33 +268,59 @@ impl AppState {
         if let Some(_entry) = traversal.tree.node_weight(index) {
             let path_to_delete = path_of(&traversal.tree, index);
             delete_directory_recursively(path_to_delete)?;
```
FedericoStra, line 270: I have a concern here. If `delete_directory_recursively` removes some nested entries but the fails, this line causes an early return. Then the subsequent line doing the cleanup of the traversal is not executed. Can this cause problems where the traversal does not match any more with the state of the filesystem?

Byron: That's an interesting spot - I also took a closer look once more to understand what's going on there.

I agree that the view of the filesystem may not match anymore with the view we have of it in memory as we will keep the entry entirely even if only the last file fails to be deleted. This is done knowing that the user should be able to try again and that the entries pane will highlight non-existing files in red when browsing them.

> Can this cause problems where the traversal does not match any more with the state of the filesystem?

No, but the current implementation might be more surprising to some.

The _more_ proper way of doing this would certainly be to traverse the in-memory data structure and delete entries one by one after their corresponding file/directory on disk was deleted. Such functionality would have to deal with non-empty directories as somebody might have added files after due last ran, probably by deleting directories recursively even if our in-memory representation thinks it is empty already.

Maybe there is more that could be done here and I am absolutely open to your ideas or an attempt to improve the situation. Maybe that could be another PR though.

Byron (changes requested):
Great work, thanks a lot!

I would have merged it already but saw that the visual indication of what to press in the mark pane is missing.

Screenshot 2021-06-30 at 07 25 28

Is this something you want to add? If not I would add it for you and merge, but didn't want to take away from the experience.

Byron (approved):
Actually I couldn't resist to just change this message myself to get a release out without delay - trashing is such a nice feature especially as its faster while allowing a change of mind later.

PR #103, opened 2021-08-18
Remove unnecessary line

benarmstead:
I highly recommend checking this, I may be wrong, but I think this line is not needed.


Byron:
Thanks! CI failed because of the formatting being incorrect though. `cargo fmt` should do it.

PR #104, opened 2021-08-18
Remove unnecessary line

benarmstead:
I think I have fixed the formatting problems faced earlier


Byron:
Thanks a lot, great catch:). One allocation less.

PR #105, opened 2021-08-22
Support Home/End and fix inconsistent help text

kidonng:
I'm one the people who use arrow keys and page up/down/home/end to navigate instead of alphabet keys 😄 and I find it weird that dua doesn't support Home/End. This is a pretty trivial change, so I hope you will accept it.

I also took the chance to fix inconsistent casing (opinionated), wording and forgotten `help_text` change to `help.rs` in 7689016c537d054a519e4e61c577e30645537213.


Byron:
Thanks a lot, the change is much appreciated.

Home and End keys aren't usually available on the Mac keyboards I am using and the arrow keys aren't anything you want to use either there ;). Being less surprising is always a welcome change.

PR #106, opened 2021-08-22
Improve mark widget tip style

kidonng:
I find the current style looking a little bit weird, so I tried to improve it a little bit, see comparison below.

Before:
image
image

After:
image
image

### DIFF
--- a/src/interactive/widgets/mark.rs
+++ b/src/interactive/widgets/mark.rs
@@ -364,7 +364,7 @@ impl MarkPane {
             Paragraph::new(Text::from(Spans::from(vec![
                 #[cfg(feature = "trash-move")]
                 Span::styled(
-                    " Ctrl + t",
+                    " Ctrl + t ",
                     Style {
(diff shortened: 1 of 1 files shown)


src/interactive/widgets/mark.rs
```diff
@@ -374,9 +374,10 @@ impl MarkPane {
                 #[cfg(feature = "trash-move")]
                 Span::styled(" to trash or ", default_style),
                 Span::styled(
-                    " Ctrl + r",
+                    " Ctrl + r ",
                     Style {
                         fg: Color::LightRed.into(),
+                        bg: Color::Black.into(),
```
kidonng, line 380: I copied this line from above, but it may conflict with the `add_modifier` style below? I don't really know what that means or how to trigger it.

Byron:
Thanks a lot for the effort! This does indeed look a lot better. The improvements you made should now be available in v2.14.6.

PR #107, opened 2021-09-17
Fix deletion process on Rust 1.55

Bobo1239:
Rust 1.55 unfortunately breaks the existing code which assumed that
ErrorKind::Other will be returned by fs::read_dir() when a file is
encountered. For now just always try deleting as if it's a file with the
possibility to add the optimization back again once the `io_error_more`
feature is stabilized.

References:
- https://blog.rust-lang.org/2021/09/09/Rust-1.55.0.html#stdioerrorkind-variants-updated
- https://github.com/rust-lang/rust/pull/85746
- https://github.com/rust-lang/rust/issues/86442

### DIFF
--- a/src/interactive/app/handlers.rs
+++ b/src/interactive/app/handlers.rs
@@ -426,13 +426,15 @@ fn delete_directory_recursively(path: PathBuf) -> Result<(), usize> {
                     }
                 }
             }
-            Err(ref e) if e.kind() == io::ErrorKind::Other => {
-                // assume file, save IOps
-                num_errors += into_error_count(fs::remove_file(path));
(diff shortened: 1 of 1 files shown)


Byron (approved):
Thanks a lot! Even though it's great to see more detailed io error codes, it's the second project that was bitten by it by surprise and I wonder if there would have been other ways to introduce this, like a deprecation period. Otherwise everybody would be expected to test against beta or nightly where this might or might not have surfaced in the first place.

But I am rambling, a new release will be out soon :).

Byron:
v2.14.7 was just released.

PR #114, opened 2021-11-09
Fix link to releases

kuviman:
One of the link to releases section is actually just linking to the repo currently


Byron:
Thanks a lot, much appreciated.

PR #117, opened 2022-01-09
fix: build on platforms without 64-bit atomics

cinerea0:
This change would allow dua to be built on platforms without 64-bit atomics.


Byron:
Thanks so much! I used this opportunity to cut a [new release](https://github.com/Byron/dua-cli/releases/tag/v2.16.0) as well, maybe that helps.

cinerea0:
Thank you for being so responsive! And yes, cutting a new release helps a lot; thank you for that as well.

PR #118, opened 2022-01-21
Add `a` key to toggle marked status of all entries

Omnikar:
Hi, I just found this project, and I think it's great! However, one inconvenience I've found with it is the inability to mark all files in a directory for deletion. I have added <kbd>a</kbd> to fulfill that functionality here.

I extracted the functionality of the `mark_entry` function that uses `index` into a function `mark_entry_by_index` that is then used in the original `mark_entry` function and a new `mark_all_entries` function.

Another thing to mention is that I made <kbd>a</kbd> toggle the marked status of all entries, rather than marking all of them, because I think the former would be slightly more useful in that if you wanted to mark all but select files, you could start by marking those select files, then toggle all.

### DIFF
--- a/src/interactive/app/eventloop.rs
+++ b/src/interactive/app/eventloop.rs
@@ -136,6 +136,7 @@ impl AppState {
                         window,
                         traversal,
                     ),
+                    Char('a') => self.mark_all_entries(MarkEntryMode::Toggle, window, traversal),
                     Char('u') | Char('h') | Backspace | Left => {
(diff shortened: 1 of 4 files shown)


Byron:
Thanks a lot, it's an awesome contribution paired with great execution. In e2686952b4daf4c35303689c36bebc3dfe3faf29
 I added more docs, but that was it.

A [new release](https://github.com/Byron/dua-cli/releases/tag/v2.17.0) was published with the change.

PR #119, opened 2022-01-23
Update clap to official release

tranzystorek-io:
Updates the clap version to the 3.0 release as well as simplifies some code and definitions


Byron:
Thanks so much for one of these rare PRs that actually remove code while making everything else more readable.

Some of these simplifications can also be applied to `gitoxide`, I definitely learned some new tricks thanks to you.

PR #129, opened 2022-06-10
show the current directory in the footer

smason:
I kept getting lost in interactive mode so thought it would be nice to display where the user currently is in the hierarchy

Not sure if this is the right place to add this change, but it seems to do something useful for me.

### DIFF
--- a/src/interactive/widgets/main.rs
+++ b/src/interactive/widgets/main.rs
@@ -55,11 +55,7 @@ impl MainWindow {
                 add_modifier: Modifier::empty(),
                 ..Style::default()
             };
-            let bold = Style {
-                fg: Color::Rgb(230, 230, 230).into(),
-                add_modifier: Modifier::BOLD,
-                ..grey
-            };
+            let bold = Style::default().add_modifier(Modifier::BOLD);
             match state.focussed {
                 Main => (bold, grey, grey),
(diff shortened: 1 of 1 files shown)


Byron (changes requested):
Thanks for your contribution! When reading the intention I thought that this would be something I want to, and when looking at it I noticed something.

Screen Shot 2022-06-10 at 21 10 02

In case it's unclear, the current path is already shown in the top-left and it is displayed in bold.

Under the circumstances, do you think displaying it again in the bottom is worth it?

smason:
ah, just noticed that it's because it's because I'm using "light" theme, so it's effectively invisible for me:

20220610_18h05m27s_grim

not sure what to suggest, maybe using colors differently help fix this and #13?

Byron:
Ah, that makes sense! Maybe you can change the color of the regular path to the one already used for paths elsewhere, that should make it visible and is the simplest solution I can think of.

What do you think?

smason:
just had a chance to play with this and a trivial change to colors looks good to me

20220614_13h01m49s_grim

(forced push because no point actually showing the path twice!)

smason:
thanks!  hopefully my use of `du -s * | sort -rn | head` will start to decrease now

PR #133, opened 2022-07-19
Add Scoop installation for Windows

sitiom:
[`dua`](https://scoop.sh/#/apps?q=main%2Fdua&s=0&d=1&o=true) is available in Scoop for Windows.


Byron:
Oh, that's cool, thanks a lot for contributing!

_(I wonder if there is a way to run it properly, in my attempt the interactive mode was quite botched in an msysgit terminal.)_

sitiom:
It works fine for me in the Windows Terminal and the Conhost. However, I cannot interact with anything in the Mintty (MSYS2 and Git Bash).

PR #136, opened 2022-07-21
Add Apple M2 to default thread derivation

### LINKED ISSUE #135: Couldn't auto-configure correct amount of threads for Apple M2
I get this message when running on my M2 Air. Just opening this issue for tracking, I'll start on a fix now:)

fmckeogh:
Closes #135 (I really overestimated how complex these changes would be 😅)

Measured 1-8 threads on both a small (100MB) and large (80GB) directories, 4 threads was the most performant in both cases.


Byron:
Thanks for testing, it's good to know the magic number 4 is still a thing :).

PR #139, opened 2022-09-01
Dependencies: switch from colored to owo-colors

murlakatamenka:
`owo-colors` is well-maintained, zero-allocation, zero-dependencies crate
for terminal colors. Also it works on any type that implements Display trait,
not just on strings.

I made minimal changes to the source code for this switch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -101,17 +101,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "colored"
-version = "2.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


Byron (changes requested):
Thanks a lot for contributing this improvement!

When running it and comparing against the baseline, I noticed the formatting seems to get lost along the way:

Screen Shot 2022-09-02 at 09 41 53

Once this is fixed, I think it's ready to be merged.

murlakatamenka:
So, there it is, sorry it took so long.

I did refactor that function, no more `.to_string()` on paths now since `owo-colors` works on anything that implements `Display` trait.

As for the the broken size column I restored `.to_string()` call on `ByteFormatDisplay`, without it the alignment breaks.

Byron:
No worries, and thanks a lot for following through.

PR #145, opened 2023-02-05
Draw window before processing events, fixes #143

### LINKED ISSUE #143: Output not refreshing since v2.18
I installed `dua` on my Linux computing cluster, which I access via `ssh` on my iTerm terminal. Version v2.17.8 runs perfectly, but all newer versions (I've tested 2.19.0 and 2.18.2) fail to actually refresh the terminal unless the user interacts using the keyboard. For example, the application opens with a blank screen, and pressing up on the keyboard will cause the TUI to show, but then the scanned files count doesn't update until I press another button on the keyboard. It simply won't refresh by itself. Did you bump one of the terminal rendering libraries or adjust the rendering behaviour with the move to v2.18.x? Happy to provide more info to help diagnose this.

fmckeogh:
Fixes #143 (on my machine). It looks like commit 28f5ac90cc1ba7d668ae8a83eb5cd899294a8301 unintentionally removed a call to `draw`. Putting it back in appears to fix the issue. Will definitely need to verify on other people's machines.

### DIFF
--- a/src/interactive/app/eventloop.rs
+++ b/src/interactive/app/eventloop.rs
@@ -77,6 +77,7 @@ impl AppState {
         use crosstermion::input::Key::*;
         use FocussedPane::*;
 
+        self.draw(window, traversal, *display, terminal)?;
         for event in events {
             let key = match event {
                 Event::Key(key) => key,


Byron:
That's a great find, thanks so much! v2.19.1 is on the way.

PR #146, opened 2023-02-22
traverse: don't recurse on cross-device filesystems

krallin:
Like it says in the title. Right now, if you pass `-x`, dua doesn't count files on other devices, but it still enumerates them. However, a good reason to use `-x` is if you have network mounts that are slow, so this fixes that.

### DIFF
--- a/src/aggregate.rs
+++ b/src/aggregate.rs
@@ -97,7 +97,7 @@ pub fn aggregate(
                 continue;
             }
         };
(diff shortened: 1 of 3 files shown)


Byron:
Thanks a lot for this small fix with huge effects for performance.

[A new release](https://github.com/Byron/dua-cli/releases/tag/v2.19.2) is now available to incorporate it.

krallin:
Thanks for the quick reply!

PR #147, opened 2023-02-22
Update help.rs

nyurik:
minor help text fix


Byron:
Thanks a lot, good catch!

PR #150, opened 2023-03-12
use binary format by default (except on macOS)

### LINKED ISSUE #85: Consider using binary (KiB, MiB) format by default.
It just makes more sense :moyai:, and it is the behavior of `ls -lh`, `free -h`, `du -h`.

tessus:
fixes #85

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -1,5 +1,5 @@
 use dua::ByteFormat as LibraryByteFormat;
-use std::path::PathBuf;
+use std::{env, path::PathBuf};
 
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot! And I learned something as well, as I wasn't aware of `std::env::consts` (but used conditional compilation for these things as a binary is built per platform anyway).

tessus:
You are welcome. Using conditional compilation is a mistake in most cases. e.g. for something like this it makes no sense.

Conditional compilation makes sense for libraries when the functions/methods are fundamentally different and are thousand of lines long.

PR #158, opened 2023-05-02
moonwalk (next major version)

Byron:
Integrate `moonwalk` for faster iteration.

### Tasks

* [x] use in aggegrate
* [x] proper size computation for roots that are files (needs `std::fs::Metadata` -> `moonwalk::Metadata`)
* [x] use in TUI
* [x] sort entries in TUI when done or as step when displaying them.
* [x] hardlink-counting
* [ ] size-on-disk for windows
* [ ] upgrade to release version of `moonwalk`

### Missing Features (for tests)

* sorting by name (how to solve this one?)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -292,6 +292,7 @@ dependencies = [
  "num_cpus",
  "open",
  "owo-colors",
+ "parking_lot",
  "petgraph",
  "pretty_assertions",
  "trash",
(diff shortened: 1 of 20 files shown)


Byron:
It's probably going to be a while until `moonwalk` is ready for consumption so we'll be back later.

PR #162, opened 2023-05-10
feat: use ratatui instead of tui

joshka:
- cargo toml patch is temporary and should be removed once the ratatui patches for crosstermion and tui-react are applied

I ran make tests

Relies on https://github.com/Byron/tui-crates/pull/2

[asciicast](https://asciinema.org/a/6bzustyRZCzquhxgpsJpMEyfj)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -225,9 +225,9 @@ dependencies = [
 
 [[package]]
 name = "crossterm"
-version = "0.25.0"
+version = "0.26.1"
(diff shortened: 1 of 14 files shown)


Byron:
Thanks for kicking this off!

The major release of `dua` is in the works [over here](https://github.com/Byron/dua-cli/pull/158) and that's where I have made the switch to `ratatui` as well, but using dependency renaming instead of renaming imports. Thanks for your understanding.

Issue #167, opened 2023-06-30
alignment off when in interactive mode and -f binary


tessus:
```
│  2.43 GiB |   0.8%  ▏         |  .gradle                                                                                                                                                     │
│1010.36 MiB |   0.3%            |  .yarn                                                                                                                                                      │
│807.69 MiB |   0.3%            |  .npm-global                                                                                                                                                 │
│412.85 MiB |   0.1%            |  .vscode                                                                                                                                                     │
```

I think when using `-f binary` the left column must be one character wider, because the max number value is 1024 and thus won't change units (start at 0 again) after 999.99.

Byron:
I have trouble reproducing it. Can you validate you are using the latest published version (v2.20.1)?

tessus:
Yep, I an sure:

<img width="488" alt="image" src="https://github.com/Byron/dua-cli/assets/223439/9137ed13-dcb5-4038-b0de-0517ce8fec4f">

QuentinMoss:
I can take a look at this one @Byron

QuentinMoss:
@Byron I'm not able to reproduce this on OS X or Pop_OS.

tessus:
It's a weird situation and not easy to reproduce. I created a single file with 1059439247 bytes in a directory. I can't reproduce it with that file either.

But with a `.yarn` directory with many folders and files totalling 1059439247 bytes, the issue manifests itself.

Here is a compressed version. I used the highest compression rate, but it's still 710MB:
https://evermeet.cx/pub/tmp/yarn.7z

Uncompress it in a subdirectory and run `dua -f binary i` in the folder above.

tessus:
Wait. A few less bytes will also do:

```
mkfile 1059438592 hello.file
```

QuentinMoss:
There's definitely some alignment off when using interactive:
<img width="2503" alt="Screenshot 2023-07-06 at 10 33 50 PM" src="https://github.com/Byron/dua-cli/assets/350409/8c6d9dfd-32fe-4a0b-a15f-e0ea7f313d6d">

tessus:
Also in non-interactive mode:

```
$ dua
  4.00 KiB .editorconfig
  4.00 KiB .gitignore
  4.00 KiB Cargo.toml
  4.00 KiB LICENSE
  4.00 KiB Makefile
  4.00 KiB etc
  8.00 KiB README.md
 12.00 KiB ci
 16.00 KiB .github
 32.00 KiB Cargo.lock
[...]

PR #169, opened 2023-07-07
Improve disk usage report formatting

QuentinMoss:
Add one space to the left column.

https://github.com/Byron/dua-cli/issues/167

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "anstream"
-version = "0.3.1"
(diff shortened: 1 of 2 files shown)


Byron (changes requested):
Thanks a lot for giving it a shot.

It looks like the change also affects the metric format though, and generally adds quite some padding now. The idea was that it only uses exactly as much space as needed, and that feature seems to be lost now.

Maybe there is another way to implement this?

Screenshot 2023-07-07 at 08 48 53

*(left: this PR, right: current version in `main`)*

Issue #34, opened 2020-02-22
Marking parent directory for deletion counts children twice


thomwiggers:
I marked `~/.cache/yay/packages` for deletion, and then decided I could get rid of all of it: selecting `~/.cache/yay/` for deletion results in dua saying it's going to delete 16 GiB, while it's actually only going to delete 9.

![image](https://user-images.githubusercontent.com/974662/75084028-4cc13680-54eb-11ea-839a-5bcd31607bd8.png)

Byron:
Thanks for posting! 

It's a known issue, which won't be a show stopper as double-deletion of files has no effect, so `Ctrl+R` still works as expected.

It's probably something that is easily fixable, too.

AngelLomeli:
@Byron Hello, I'm interested in taking a look at this issue. I stumbled across the project earlier today and I'm liking it so far - it has already been much faster than waiting on standard `du`! I'm fairly new to Rust, but experienced in several other languages.

I see that the calculation is being made in `MarkPane::render` with the following line. My first impression is that the iterator could be filtered to exclude any EntryMarks that represent a directory that exists in a subdirectory of another EntryMark.
`format.display(marked.iter().map(|(_k, v)| v.size).sum::<u64>())`

The naive approach would seem to be another iteration in that filter over all of the EntryMarks and canonicalizing the paths to account for symlinks. My concern with that approach is that it turns an O(N) iteration into an O(N^2) operation in the render loop. If the EntryMarkMap is relatively small, that may not be a major issue.

Would you say I'm on the right track here, or would the better approach be to prevent entries from being marked if they already have a parent directory in the EntryMarkMap? That would move the more expensive calculation outside the render loop, but would raise some additional questions about the desired behavior.

I'll try to get some actual code in place in the morning when I've had a bit more time to look through the existing code and get a better feel for it.

Thanks!
Angel Lomeli

Byron:
Hi @AngelLomeli ,

so great to hear you would like to give it a try, and here you will receive nothing but encouragement in that undertaking :)!

You are right, applying the filter at render time might be doing unnecessary work, and maybe doing it when the entry is added is a better choice. Then even the naive, slow approach might be absolutely good enough.

A good place to get started is to check what happens when a key is pressed - the [event loop is where it all happens](https://github.com/Byron/dua-cli/blob/cb2bbdfe616b38311ebe26e78999c69a4637a5dd/src/interactive/app/eventloop.rs#L111).

[...]

PR #178, opened 2023-11-20
Fixes marking parent directory for deletion counts children twice

gosuwachu:
Issue #34 

Screenshot 2023-11-20 at 22 10 52

### DIFF
--- a/src/interactive/widgets/mark.rs
+++ b/src/interactive/widgets/mark.rs
@@ -33,6 +33,8 @@ pub enum MarkMode {
 }
 
 pub type EntryMarkMap = BTreeMap<TreeIndex, EntryMark>;
+
+#[derive(Default)]
 pub struct EntryMark {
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot for contributing a fix for this long-standing issue!

I won't create a release of this right away in the hopes you have plans for additional fixes. If not, please let me know and I can create another follow-up release as well.

gosuwachu:
Thanks @Byron! I am not planning on fixing more issues, but I may write more tests in similar fashion to the tests in this PR. I am doing this for fun / to practice programming in Rust a bit.

Byron:
That's great to hear! A few more tests definitely won't hurt. Also note that there is a way to write journey tests, and currently there isn't many user-flows under test.

In any case, [here is the new release](https://github.com/Byron/dua-cli/releases/tag/v2.20.3).

Issue #141, opened 2022-12-02
Support sorting by mtime


SUPERCILEX:
It'd be nice to be able to sort by oldest modified files/directories.

Byron:
Could you define this enhancement more so it could be implemented by a contributor? Where does it apply? Which flags to use, etc… . Thanks

SUPERCILEX:
This is in interactive mode. Instead of just being able to sort by file/directory size, you should be able to sort by last modified time. Ncdu uses the M shortcut: https://dev.yorhel.nl/ncdu/man2#:~:text=Order%20by%20latest%20child%20mtime%2C%20or%20modified%20time.%20(press%20again%20for%20descending%20order)%20Requires%20the%20%2De%20flag.

Byron:
Good catch!

PR #179, opened 2023-11-23
Adds keybinding for 'm' to toggle sorting by modified time

gosuwachu:
Adds keybinding for 'm' to toggle sorting by modified time. Additionally appropriate column is highlighted depending on if sorting is by time or by size + mtime column is hidden by default.

Sorting by modified time + current sorting mode in the footer:
Screenshot 2023-11-23 at 22 59 52

Help:
Screenshot 2023-11-23 at 23 00 08

Addresses issue #141

### DIFF
--- a/src/traverse.rs
+++ b/src/traverse.rs
@@ -266,7 +266,7 @@ mod tests {
     fn size_of_entry_data() {
         assert_eq!(
             std::mem::size_of::<EntryData>(),
-            if cfg!(windows) { 72 } else { 64 },
+            if cfg!(windows) { 64 } else { 64 },
             "the size of this should not change unexpectedly as it affects overall memory consumption"
         );
     }
--- a/src/traverse.rs
+++ b/src/traverse.rs
@@ -3,22 +3,46 @@ use anyhow::Result;
 use filesize::PathExt;
 use petgraph::{graph::NodeIndex, stable_graph::StableGraph, Directed, Direction};
 use std::{
+    fmt,
     fs::Metadata,
     io,
     path::{Path, PathBuf},
-    time::Duration,
+    time::{Duration, SystemTime, UNIX_EPOCH},
 };
 
 pub type TreeIndex = NodeIndex;
 pub type Tree = StableGraph<EntryData, (), Directed>;
 
-#[derive(Eq, PartialEq, Debug, Default, Clone)]
+#[derive(Eq, PartialEq, Clone)]
 pub struct EntryData {
     pub name: PathBuf,
     /// The entry's size in bytes. If it's a directory, the size is the aggregated file size of all children
     pub size: u128,
+    pub mtime: SystemTime,
     /// If set, the item meta-data could not be obtained
     pub metadata_io_error: bool,
+}
+
+impl EntryData {
+    pub fn default() -> EntryData {
+        EntryData {
+            name: PathBuf::default(),
+            size: u128::default(),
+            mtime: UNIX_EPOCH,
+            metadata_io_error: bool::default(),
+        }
+    }
+}
+
+impl fmt::Debug for EntryData {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("EntryData")
+            .field("name", &self.name)
+            .field("size", &self.size)
+            // Skip mtime
+            .field("metadata_io_error", &self.metadata_io_error)
+            .finish()
+    }
 }
 
 /// The result of the previous filesystem traversal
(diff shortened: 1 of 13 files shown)


src/traverse.rs
```diff
@@ -3,24 +3,48 @@ use anyhow::Result;
...
 pub type TreeIndex = NodeIndex;
 pub type Tree = StableGraph<EntryData, (), Directed>;
 
-#[derive(Eq, PartialEq, Debug, Default, Clone)]
+#[derive(Eq, PartialEq, Clone)]
 pub struct EntryData {
     pub name: PathBuf,
     /// The entry's size in bytes. If it's a directory, the size is the aggregated file size of all children
     pub size: u128,
+    pub mtime: SystemTime,
```
Byron, line 21: Could we have a test along the lines of `assert_eq!(std::mem::size_of::<EntryData>(), X)` to see what the size of this is now?

You could just rebase onto `main` to get such a test, I added it. I'd not expect this to have huge impact, but I'd like to be sure. Probably it's best to run both versions, with and without mtime, on a big tree and compare the consumption for completeness.

Byron (changes requested):
Thanks a lot for adding such a huge feature - t's much appreciated!

I'd expect the performance to remain roughly the same as metadata is already queried anyway. The only difference should be memory consumption as the mtime field has been added.

Besides that, I'd love to retain the initial cleanness of the GUI, and start with the `mtime` column hidden. This can probably be implemented by adding another state to the sorting toggle.

* initial state: mtime hidden
* press m: sort by mtime ascending
* press m: sort by mtime descending
* press m: mtime hidden

Does that make sense?

Thanks for your consideration.

src/traverse.rs
gosuwachu: I have just rebased, and re-runt the tests:

```
---- traverse::tests::size_of_entry_data stdout ----
thread 'traverse::tests::size_of_entry_data' panicked at 'assertion failed: `(left == right)`
  left: `64`,
 right: `48`: the size of this should not change unexpectedly as it affects overall memory consumption', src/traverse.rs:267:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Looks like mtime adds 16bytes per entry.

gosuwachu: Let me know if this is too much in your opinion, and I will replace it with a single u64 for unix timestamp instead (8 bytes).

Byron: Yes, that's a good idea. Let's see if a single u64 would actually safe 8 bytes, but if so I think we should have it. Nanosecond precision for mtimes isn't needed, after all.

> [!NOTE]
> If you are inclined to have nanoseconds, you can also use an Ext4-style encoding of 30 bits of nanoseconds and 32 bits of seconds, along with 2 bits of epoch. But that's entirely optional and to my mind, unnecessary (*I just like the encoding and wanted to have mentioned it 😁*)

Byron (changes requested):
Thanks a lot for applying the requested changes.

I left a note regarding the size of the structure, if you could have a look.

Further, I ran the program and love that the mtime column now starts hidden. However, I'd expect it to cycle through the hidden state as well, and right now there is no way to get to the original, hidden, state anymore. Was this intentional, and if so what's the rationale?

Thanks again for your help in getting this feature implemented.

gosuwachu:
The way to hide the mtime column is to sort by size. You press 'm' to sort by mtime, and press 's' to sort by size. If you keep pressing 'm' or 's' you just cycle between ascending / descending. My rationale is that if the user wants to sort by mtime then that column should be visible.

src/traverse.rs
gosuwachu: I have made the change, but it has not saved any memory, because of struct memory alignment. The largest member in `EntryData` is 8 bytes (`size` field), so the other fields are aligned to the same size (https://levelup.gitconnected.com/how-struct-memory-alignment-works-in-c-3ee897697236).

What do you think? Given the above problem, using `SystemTime` actually makes the code cleaner, as otherwise I have to deal with the problem of converting timestamp back to a `DateTime` and formatting it.

Byron: I had a hunch this could happen as well. Let's keep it then.

Byron:
Thanks for the explanation, that also makes sense and I think it's enough to wait for feedback from users. My guess is that it can stay as is as well.

Screenshot 2023-11-25 at 14 42 36

The above is a screenshot of `dua` after seeing a lot of files. 

After trying the same with this branch, I got a crash (*wait for it :D*):

```
└──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── mark-move = d | mark-toggle = space | toggle-all = a┘
Sort mode: size descending  Total disk usage: -  Entries: 772604 in 5s (145882thread 'main' panicked at library/std/src/sys/unix/time.rs:77:9:
                                                                                                                                              assertion failed: tv_nsec >= 0 && tv_nsec < NSEC_PER_SEC as i
64
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
                                                                               [1]    49031 abort      cargo run --release -- i ~/dev
                                                                                                                                     %
```

It crashes in the standard library which isn't able to deal with *timestamps before 1970 with nanoseconds on MacOS*. Yes, it's true, and it will panic then.

Fortunately, [this was already fixed](https://github.com/rust-lang/rust/pull/117451) and will be available in Rust 1.75. This should just mean I have to hold back with creating a new release until then.

Let's hope I remember.

gosuwachu:
> Thanks for the explanation, that also makes sense and I think it's enough to wait for feedback from users. My guess is that it can stay as is as well.
> 
> Screenshot 2023-11-25 at 14 42 36
> The above is a screenshot of `dua` after seeing a lot of files.
> 
> After trying the same with this branch, I got a crash (_wait for it :D_):
> 
> ```
> └──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── mark-move = d | mark-toggle = space | toggle-all = a┘
> Sort mode: size descending  Total disk usage: -  Entries: 772604 in 5s (145882thread 'main' panicked at library/std/src/sys/unix/time.rs:77:9:
>                                                                                                                                               assertion failed: tv_nsec >= 0 && tv_nsec < NSEC_PER_SEC as i
> 64
>   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
>                                                                                [1]    49031 abort      cargo run --release -- i ~/dev
>                                                                                                                                      %
> ```
> 
> It crashes in the standard library which isn't able to deal with _timestamps before 1970 with nanoseconds on MacOS_. Yes, it's true, and it will panic then.
> 
> Fortunately, [this was already fixed](https://github.com/rust-lang/rust/pull/117451) and will be available in Rust 1.75. This should just mean I have to hold back with creating a new release until then.
> 
> Let's hope I remember.
> 
> Anyway, a nightly build produced a working binary on MacOS, and here is the outcome: Screenshot 2023-11-25 at 14 57 22
> 
> Thus nothing changed beyond a rounding error, both in terms of performance or in terms of memory consumption.
> 
> Good to go 😊👍

Nice find! I have learned something new, thanks for sharing.

I know this PR has already landed, but let's keep an option of reverting it and keeping it on a branch until Rust 1.75 is released.

PR #185, opened 2023-12-05
Increase optimization level

gosuwachu:
I know this is super low hanging fruit so I am probably missing something, but why was it set to `0`?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ incremental = false
 overflow-checks = false
 lto = "fat"
 codegen-units = 1
-build-override = { opt-level = 0 }
+build-override = { opt-level = 3 }
 
 [dev-dependencies]
 pretty_assertions = "1.0.0"


gosuwachu:
> 16e00de Speedup build times by not optimizing build dependencies

I think this project is now at a point where you probably want to increase runtime performance :)

Byron (approved):
Yes, indeed, it's about time! Great catch :D.

In practice, the difference on big(ish) directories is not measurable, and with small repositories it's a little faster but those 0.5ms aren't going to relevant to any human.

```
❯ hyperfine -w1 target/release/dua dua  -N
Benchmark 1: target/release/dua
  Time (mean ± σ):       7.6 ms ±   0.4 ms    [User: 7.0 ms, System: 26.0 ms]
  Range (min … max):     6.8 ms …   9.6 ms    312 runs

Benchmark 2: dua
  Time (mean ± σ):       8.2 ms ±   0.4 ms    [User: 7.1 ms, System: 26.1 ms]
  Range (min … max):     7.5 ms …   9.1 ms    352 runs

Summary
  target/release/dua ran
    1.08 ± 0.08 times faster than dua

dua-cli ( increase-optimzation-level) [$] took 5s
❯ hyperfine -w1 'target/release/dua ~/dev/github.com/WebKit' 'dua ~/dev/github.com/WebKit'
Benchmark 1: target/release/dua ~/dev/github.com/WebKit
  Time (mean ± σ):      2.359 s ±  0.031 s    [User: 0.668 s, System: 16.249 s]
  Range (min … max):    2.328 s …  2.422 s    10 runs

Benchmark 2: dua ~/dev/github.com/WebKit
  Time (mean ± σ):      2.347 s ±  0.006 s    [User: 0.668 s, System: 16.460 s]
  Range (min … max):    2.339 s …  2.361 s    10 runs

Summary
  dua ~/dev/github.com/WebKit ran
    1.00 ± 0.01 times faster than target/release/dua ~/dev/github.com/WebKit
```

The compile times change from 22s (-O0) to 24s (-O3), which I think is negligible.


On the basis on the slightly improved startup times I think this should be merged.

Issue #110, opened 2021-10-24
Skip through single parent directory in interactive mode


kidonng:
If given a single directory as parameter in interactive mode (e.g. `dua i Developer`), dua-cli's initial state will be like:

<img width="831" alt="Screen Shot 2021-10-25 at 01 33 57" src="https://user-images.githubusercontent.com/44045911/138605860-568a0c6f-7cce-4c73-90e7-4bdbf26b981d.png">

However there's nothing useful here, the total usage is available at the bottom and the percentage will certainly be 100%. What users expect is more likely this (e.g. `dua i Developer/*`):

<img width="831" alt="Screen Shot 2021-10-25 at 01 33 52" src="https://user-images.githubusercontent.com/44045911/138605877-fc40e869-c3c7-4c34-be3a-dd40e3ea27d6.png">

(The entries difference is because we are not counting the parent directory, but that should not matter.)

There should a badge or something like that to inform the users dua-cli has skipped through the parent directory.

---

If dua-cli can skip all single parent directory, that will be more awesome.

Let's suppose we have the following file structure:

```
a
└── b
   └── c
      ├── 1
      ├── 2
      ├── 3
      ├── 4
      ├── 5
      ├── 6
      ├── 7
      ├── 8
      ├── 9
      └── 10
```

It is ideal that `dua i a` equals `dua i a/b/c/*`.

gosuwachu:
IMHO this is as much done as it makes sense to do.

Byron:
Thanks for the reminder!

PR #186, opened 2023-12-06
Skip through single root directory

gosuwachu:
Implements the suggestion from #110 to skip single root directory.

Before:
Screenshot 2023-12-06 at 00 05 25

After:
Screenshot 2023-12-06 at 00 05 37


Byron (commented):
Thanks a lot for tackling this!

I made a fix before merging if you want to take a look.

PR #188, opened 2023-12-08
Adds keybinding for 'c' to toggle sorting by number of items

gosuwachu:
Adds keybinding for 'c' to toggle sorting by number of items. I have used `human_format` to convert numbers to more human readable form.

Number of items in each directory is collected the same way as the size was done before during file system traversal. I have updated the tests to ensure the counts are calculated correctly.

Sorting by number of items + current sorting mode in the footer:
Screenshot 2023-12-08 at 22 16 47

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -277,6 +277,7 @@ dependencies = [
  "clap",
  "crosstermion",
  "filesize",
+ "human_format",
  "itertools",
  "jwalk",
  "num_cpus",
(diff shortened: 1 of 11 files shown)


Byron (approved):
Thanks a lot, it's a very nice feature to have!

I have made a few changes if you want to take a look.

Memory consumption doesn't change significantly compared to the version without counts, so tracking it seems mostly free.

PR #189, opened 2023-12-09
Displays total item count

gosuwachu:
Before:
Screenshot 2023-12-09 at 14 19 59

After:
Screenshot 2023-12-09 at 14 19 49

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -281,6 +281,7 @@ dependencies = [
  "itertools",
  "jwalk",
  "num_cpus",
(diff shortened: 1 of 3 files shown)


Byron (approved):
Thanks a lot, that's a great idea!

This inspired me to also add the aggregate size of the hierarchical level, which now looks like this:

Screenshot 2023-12-09 at 19 02 55

gosuwachu:
Nice! Looks great.

PR #190, opened 2023-12-09
Adds new target that runs all the checks (to be used before creating a PR)

gosuwachu:


Byron:
That's a good idea, thanks!

Issue #154, opened 2023-04-20
Feature: allow glob-based recursive interactive search


nyurik:
As a user, in interactive mode, I would like to find all `**/target` directories and mark them for deletion.

Byron:
Sometimes I also wish for having this feature.

For implementing it that would definitely need some fleshing out, but that can probably be done by the one picking it up.

gosuwachu:
I think better option may be to let the user run a command and use its output to the mark files.

It would be a considerable effort to implement fully featured search, and even then `dua` would be competing with tools better designed to do it.

E.g. you can run:
```
$ find . -name "target" -type d
./closure-mut/target
./alacritty/target
./dua-cli/target
./rust-hello/target
```

nyurik:
@gosuwachu globs tend to be far easier than remembering the complex syntax of the `find` or `fd` commands.  If someone is proficient with the CLI tools and remembers all the flags by heart, they are less likely to use an interactive tool to do the same. On the other hand, many users of a rich UI tool are less likely to remember the CLI command parameters.

So yes, I see value in arbitrary list, e.g. supplying a file with a list of files or dynamically generating it from a command output, but I think the lower hanging fruit that would cover 80% would be to support a simple glob pattern.

Byron:
I agree that integrating simple globs would be the way to go here, and think that the [`glob`](https://crates.io/crates/glob) crate would do the job wonderfully.
There is also https://github.com/helix-editor/nucleo , which might be an even better choice as it's optimized to apply a single expression to a whole set of items/paths.

I think the UX of this feature would also be interesting to flash out.

I imagine it to be invoked with `/`, which automatically activates the mark pane, which receives all matches as the user types an expression. The expression will be applied to all paths reachable from the current visible folder in the entries list. Some cleanup should be applied to display the shortest viable path, i.e. `target` should match `target/` and `foo/target`, but not everything within `target/`. It's also the question what's better here, apply the search to a flattened list of the directory tree, or to leverage the hierarchy. Maybe an iterative approach is possible, too, and one can start on this level of the hierarchy, use matches as is, while going deeper only where there is no match in the hope to find one then. For this case, I believe `glob` would be better, which then matches one expression against one path at a time.

Definitely an interesting problem to solve and to experiment with.

Byron:
Here is a proposition for the UX side of things:

* Press `/` to activate the mark pane and display a prompt line
     - below the yellow warning line ("hit Ctrl + r to delete") if it's on top
[...]

PR #192, opened 2023-12-10
Refactors MainWindow render to make it more readable

gosuwachu:
In preparation to add UI for #154.


Byron (approved):
Thanks a lot!

It's exciting to see you are going to tackle glob-based marking as well!

I'd appreciate if we could align on the 'UX' side of things beforehand if possible. My vision if this has become quite clear in the past couple of hours or days, but maybe there are versions of it that I didn't see. I will write more [underneath the previous ramblings](https://github.com/Byron/dua-cli/issues/154#issuecomment-1848668852) in a bit.

PR #193, opened 2023-12-10
Calculates mark pane item count consistently with the rest of the app

gosuwachu:
Screenshot 2023-12-10 at 01 07 33


Byron (approved):
Yes, that makes sense and is so much better in estimating how long the deletion is going to take.

Thanks a lot!

I unified the use of `COUNT`, also to indicate that count display should indeed be unified across the whole application.

PR #198, opened 2023-12-11
fix: EntryData size test passes on 32-bit (#197)

### LINKED ISSUE #197: Test fails on 32-bit targets
When running tests on an i686 target, the following failure occurs:

```
---- traverse::tests::size_of_entry_data stdout ----
thread 'traverse::tests::size_of_entry_data' panicked at src/traverse.rs:315:9:
assertion `left == right` failed: the size of this should not change unexpectedly as it affects overall memory consumption
  left: 56
 right: 80
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This failure can be seen in CI [here](https://github.com/void-linux/void-packages/actions/runs/7171417083/job/19526367908?pr=47626).

Reading the assertion message, is it necessary that the size of the data be exactly 80, or just that it be less than or equal to 80?

cinerea0:
This allows the EntryData size test to pass on 32-bit targets. This has been proven to work in CI here: https://github.com/void-linux/void-packages/pull/47626.

There are crates like [assertables](https://docs.rs/assertables/latest/assertables/) and [more_asserts](https://docs.rs/more-asserts/latest/more_asserts/) that offer an `assert_eq!`-like interface for less-than-or-equal-to comparisons if you'd prefer that, but I didn't want to add a dependency on your behalf.

---

Closes #197.

### DIFF
--- a/src/traverse.rs
+++ b/src/traverse.rs
@@ -312,10 +312,10 @@ mod tests {
 
     #[test]
     fn size_of_entry_data() {
-        assert_eq!(
-            std::mem::size_of::<EntryData>(),
-            80,
-            "the size of this should not change unexpectedly as it affects overall memory consumption"
+        assert!(
+            std::mem::size_of::<EntryData>() <= 80,
+            "the size of this ({}) should not exceed 80 as it affects overall memory consumption",
+            std::mem::size_of::<EntryData>()
         );
     }
 }


Byron:
Thanks a lot for your help!

In future, if you feel like adding dev-dependencies that you made good experiences with, please feel free to do so. I am always looking for new crates that make testing better.

PR #199, opened 2023-12-12
Refactors AppState initialisation during app startup

gosuwachu:
I have found previous version with the `Option<>` a bit weird, but it is probably a matter of taste. In the new version AppState is created at the start,  but unfortunately we have to deal with incoming events which complicate things.

However, I am not sure if we need to process the events during the traversal. The app is not that usable in that state anyway - events are all processed in batches which causes the cursor jump around a bit, and the experience is not smooth at all.

Also, as an example `ncdu` doesn't allow any interaction during scanning. Thoughts?

### DIFF
--- a/src/interactive/app/eventloop.rs
+++ b/src/interactive/app/eventloop.rs
@@ -251,39 +251,25 @@ impl TerminalApp {
             keys
         };
 
-        let mut state = None::<AppState>;
+        let mut state = AppState {
+            is_scanning: true,
+            ..Default::default()
+        };
         let mut received_events = false;
         let traversal = Traversal::from_walk(options, input_paths, |traversal| {
-            let s = match state.as_mut() {
-                Some(s) => {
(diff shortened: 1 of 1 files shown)


Byron:
Thanks a lot, this looks so much better (*and* reduces the amount of code).

> However, I am not sure if we need to process the events during the traversal. The app is not that usable in that state anyway - events are all processed in batches which causes the cursor jump around a bit, and the experience is not smooth at all.

It's an important feature, which also allows the user to `q`uit out of the application early.
Even though it's not the most responsive right now, it's predictable, and I use early browsing all the time, even if only to keep me busy while waiting.
So this is definitely a keeper.

I may add that during my [experiments with `moonwalk`](https://github.com/Byron/dua-cli/pull/158) I noticed that it's much more responsive with it. It will just take some time until the author completes it, but it will land eventually for performance and responsiveness.

PR #200, opened 2023-12-12
Adds the missing '/' prefix for root directories

gosuwachu:
Before:
Screenshot 2023-12-12 at 23 22 05

After:
Screenshot 2023-12-12 at 23 22 14


Byron (approved):
Thanks a lot, a great catch!

Not showing the prefix `/` on the root is fully intentional, even though it now doesn't work as it was supposed to anymore. The general idea is that path given by the user are left verbatim.

I re-added some logic that will detect typical user-specified paths, but not all of them due to ambiguity.

Of course, there would be a way to do it perfectly, but I thought it wasn't worth it.

PR #205, opened 2023-12-26
Keep consistent language/punctuation/case throughout the app.

gosuwachu:
Consistent language is important. We have started using "items" in few places and it seems more correct in `dua`'s context: entry is something that has been put in vs an item in a list. I have also made sure all sentences start a with capital letter and end with a dot - it looks more mature / professional. 

Before:
Screenshot 2023-12-26 at 23 09 33

After:
Screenshot 2023-12-26 at 23 08 43

### DIFF
--- a/src/interactive/widgets/help.rs
+++ b/src/interactive/widgets/help.rs
@@ -106,105 +106,109 @@ impl HelpPane {
                 }
             };
 
-            title("Keys for pane control");
+            title("Pane control");
             {
                 hotkey(
                     "q/<Esc>",
-                    "Close the current pane. Closes the program if no",
-                    Some("pane is open"),
+                    "Close the current pane.",
+                    Some("Closes the program if no pane is open."),
                 );
                 hotkey(
                     "<Tab>",
                     "Cycle between all open panes.",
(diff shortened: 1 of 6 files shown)


src/interactive/widgets/help.rs
```diff
@@ -106,105 +106,109 @@ impl HelpPane {
...
-                hotkey("<Space>", "Toggle the currently selected entry", None);
-                hotkey("a", "Toggle all entries", None);
+                hotkey("<Space>", "Toggle the currently selected item.", None);
+                hotkey("a", "Toggle all items.", None);
                 hotkey(
                     "/",
-                    "Git-style glob search, case-insensitive, always from the top of the tree",
-                    None,
+                    "Git-style glob search, case-insensitive.",
+                    Some("Search starts from the current directory."),
```
Byron, line 182: A great catch! My testing must have fooled me into thinking otherwise.

Byron (approved):
Thanks a lot, it's a great idea to tackle this.

While working on it recently I also noticed my own insecurity about the right choice of words, but ended up ignoring it.

PR #206, opened 2023-12-26
Option to enable debug logs

gosuwachu:
I find logs very useful. So far I had to keep applying this patch on top of my changes. It would be useful to add this option to the cli to help with debugging. I hope the logs will improve in the future to include more interesting info abouy the operation of the program.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,21 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.21.0"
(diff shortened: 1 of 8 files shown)


Byron (approved):
Thanks a lot, a great idea!

The bare-bones feature didn't seem too useful, but it turns out there is a wonderful `log-panics` crate which works as advertised. Seeing panics properly fixes a serious issue I had in the rare occasion that `dua i` did crash on me - now the panic will just be written into the log file, with backtrace even.

PR #209, opened 2024-01-05
Optimize UI responsiveness in scanning state

unixzii:
The UI is very laggy while the scanning is in progress, and it makes the program almost unusable during the first few seconds of startup. The bottleneck is `jwalk` iterator, it may block the UI thread sometimes. In this PR, I alleviated the situation by some optimizations, and the UI is now super responsive at startup.

Here is what it looks like:

https://github.com/Byron/dua-cli/assets/3109144/3b06b888-c4ec-4c45-9468-36b0059cbe6c

Some implementation notes:

- The traversal operation is moved to another thread to prevent the UI thread from being blocked, and its results are sent to the UI thread via a channel.
- Use `crossbeam::channel` to receive UI events and `jwalk` entries. In contract to `std::sync::mpsc`, we can select from a set of channels without busy polling using `try_recv`.
- Created our own version of `input_channel`. The function provided by `crosstermion` returns a `std::sync::mpsc::Receiver` value, which is not compatible with `crossbeam::select!`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -356,6 +356,8 @@ dependencies = [
  "byte-unit",
  "chrono",
  "clap",
+ "crossbeam",
+ "crossterm",
  "crosstermion",
  "filesize",
  "gix-glob",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -357,7 +357,6 @@ dependencies = [
  "chrono",
  "clap",
  "crossbeam",
- "crossterm",
  "crosstermion",
  "filesize",
  "gix-glob",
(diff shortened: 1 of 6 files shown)


gosuwachu:
I have tested this change and indeed it is now super smooth 🎉  Thanks!

Byron:
Thanks a million, it works just as advertised!

It's so good to have this issue finally fixed!

And because we had to wait so long for this to be improved, I didn't want anybody to wait even longer: here is [the new release](https://github.com/Byron/dua-cli/releases/tag/v2.26.0)

PR #214, opened 2024-01-08
Refactors app initialisation 

gosuwachu:
This PR refactors app initialisation so that:
1. the app starts with an empty `Traversal`.
2. the app then starts the traversal by creating a `RunningTraversal`.
3. `Traversal` gets updated by processing events from the `RunningTraversal` inside the main event loop.

This should enable in future to implement refresh logic by creating another `RunningTraversal` and updating parts of the tree (TBD how to exactly do that).

Other than the above `AppState` and `TerminalApp` have been moved to separate files.


Byron (approved):
Thanks a lot for making this happen - I think it's a great base for implementing refresh later.

Byron:
I really like that the application will be able to trigger the refresh itself by creating a new `BackgroundTraversal` with enough information to know how to `integrate_traversal_events` in order to insert entries under a new root. Refresh should be no problem with that and fall right into place.

It also seems that `TerminalApp` doesn't fulfil a purpose anymore and that its fields can be merged into `AppState` without losing anything. This might then be helpful when initiating a new walk (as it requires `WalkOptions`). But I leave that to you :).

gosuwachu:
> It also seems that TerminalApp doesn't fulfil a purpose anymore and that its fields can be merged into AppState without losing anything. This might then be helpful when initiating a new walk (as it requires WalkOptions). But I leave that to you :).

I have thought the same thing, and was the first thing that I have tried to do when working on this change, however I have ran into ownership / borrow checker issues. I may give it another go.

gosuwachu:
Ups... all my messy commit "wip" history ended up being merged into main - I don't mind if you squash when landing the PR, but I will also try to remember to squash all the commits before sending stuff for review.

Byron:
No worries about it - I was aware and chose to prefer the detailed step-by-step history over a single commit that hides the steps. Even though I appreciate more structure, it's by no means a requirement and … I have been doing the same for a long time.

[StackedGit](https://stacked-git.github.io) changed that to the better, but it comes with a little overhead that to me at least is worth it.

Issue #96, opened 2021-06-27
Enhancement request: refresh/recompute


FedericoStra:
Hello again,

here is another idea for a potentially useful feature: add the possibility to refresh the entry list and recompute the sizes and percentages. This could be bound to the key `r`.

This is useful if the files are modified (added, removed, resized) outside of the application while it is running in interactive mode.

Byron:
That would certainly be nice to have!

Implementing it _well_ is something that wouldn't come easily to me as…

* it should be affecting only the directory tree one is currently displaying
* it should be interactive
* it should handle multiple refresh requests gracefully and properly

Instead of this vision of the perfect implementation of it maybe there is a cheaper way.

What about adding a command-line flag that allows to pass the 'root' directory at which the GUI should start. A refresh request could then `exec` dua with that flag to relaunch the GUI with the previous roots but display at the given one. While it's coming up refresh is disabled, but besides that it's just as if you would have relaunched the GUI while jumping to the target directory as fast as possible.

What do you think?

austinbutler:
The way I end up using this tool mostly is `dua i /some/path` (so it already has the path). It computes the space, then I may end up using an external tool to delete some files. I'm in a descendent of the original path I passed to `dua` and I want to press `r` and recompute the current directory recursively and without waiting for everything else to have to recompute. If re-execing dua accomplishes that it seems ok.

Byron:
> The way I end up using this tool mostly is dua i /some/path (so it already has the path). It computes the space, then I may end up using an external tool to delete some files. 

I hope you find it useful that it now prints the paths that were selected when exiting.

LoganDark:
> What about adding a command-line flag that allows to pass the 'root' directory at which the GUI should start. A refresh request could then `exec` dua with that flag to relaunch the GUI with the previous roots but display at the given one. While it's coming up refresh is disabled, but besides that it's just as if you would have relaunched the GUI while jumping to the target directory as fast as possible.

Wouldn't that cause it to forget everything else? (apologies, I am late.)

ww7:
This feature, probably most wanted from 2021

gosuwachu:
I wonder what is the use case for this feature. I get that if you delete the files outside of `dua` then that feature would be useful, but `dua` already supports deleting items from inside the program. So are users not aware that this is possible?

ww7:
@gosuwachu
[...]

PR #216, opened 2024-01-10
refresh (PoC)

Byron:
Please see the commit messages for additional context.

I won't work on this, but it should show it this can work in general,
and to my mind this is already is a successful PoC, but needs work to
deal with the details.

PS: It's also unclear how well this will work with the `glob` navigation pane, but it's probably worth trying and maybe we are lucky. If not, ignoring these keys in the glob pane should do the trick.

### DIFF
--- a/src/interactive/app/eventloop.rs
+++ b/src/interactive/app/eventloop.rs
@@ -270,6 +270,8 @@ impl AppState {
                     Char('o') | Char('l') | Enter | Right => {
                         self.enter_node_with_traversal(&tree_view)
                     }
+                    Char('R') => self.refresh(&mut tree_view, Refresh::AllInView)?,
+                    Char('r') => self.refresh(&mut tree_view, Refresh::Selected)?,
                     Char('H') | Home => self.change_entry_selection(CursorDirection::ToTop),
                     Char('G') | End => self.change_entry_selection(CursorDirection::ToBottom),
                     PageUp => self.change_entry_selection(CursorDirection::PageUp),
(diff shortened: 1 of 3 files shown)


gosuwachu:
This is as easy I as was hoping it is going to be for a simple "big-bang refresh". I will have a look what needs to be done to finish this. 

Next step after that is to implement incremental updates or at least show what part of the tree is currently being updated.

Byron:
> This is as easy I as was hoping it is going to be for a simple "big-bang refresh"

I think it's closer to [`live-refresh`](https://github.com/Byron/dua-cli/issues/96#issuecomment-1879279393) as the GUI is interactive, and it shows `scanning` as well. In theory, you *would* see things pop up as they are added to the tree, it's live after all, but the `jwalk`-powered traversal order doesn't support it, *I think*. Or in other words, if the traversal would return the top-level folder first, one would see it immediately and then its children as they appear. The traversal is already unordered as well.

Maybe tweaks to this are possible though, it seems a bit odd even though I always liked it that folders are completely known once they popup (and it would be nice to have a choice as well, like an immediate mode).

> Next step after that is to implement incremental updates or at least show what part of the tree is currently being updated.

Yes, that would be optimal, and is probably going to introduce quite some complexity, maybe more than it's worth. But we will see.

gosuwachu:
fyi, I have found quite a lot of bugs while working on this change so it will take some time before this will be ready

Byron:
No worries, no need to rush, and thanks again for your incredible help with this and `dua` in general!

gosuwachu:
Done: https://github.com/Byron/dua-cli/pull/219 This PoC can be now closed imho.

PR #218, opened 2024-01-12
code optimize

Roxannej:


src/interactive/widgets/footer.rs
```diff
@@ -25,6 +25,7 @@ pub struct FooterProps {
 
 impl Footer {
     pub fn render(&self, props: impl Borrow<FooterProps>, area: Rect, buf: &mut Buffer) {
+        use SortMode::*;
```
gosuwachu, line 28:
```suggestion
        use SortMode as E;
```

src/interactive/widgets/footer.rs
```diff
@@ -39,12 +40,12 @@ impl Footer {
...
-                    SortMode::MTimeAscending => "modified ascending",
-                    SortMode::MTimeDescending => "modified descending",
-                    SortMode::CountAscending => "items ascending",
-                    SortMode::CountDescending => "items descending",
+                    SizeAscending => "size ascending",
+                    SizeDescending => "size descending",
+                    MTimeAscending => "modified ascending",
+                    MTimeDescending => "modified descending",
+                    CountAscending => "items ascending",
+                    CountDescending => "items descending",
```
gosuwachu, line 48: This unfortunately is not recommended because it throws away some of the safety features of enums inside match expressions. See: https://www.youtube.com/watch?v=8j_FbjiowvE

```suggestion
                    E::SizeAscending => "size ascending",
                    E::SizeDescending => "size descending",
                    E::MTimeAscending => "modified ascending",
                    E::MTimeDescending => "modified descending",
                    E::CountAscending => "items ascending",
                    E::CountDescending => "items descending",
```

Byron: Thanks for this hint! I wasn't really seeing it this way just yet, but totally see the point (i.e. as enums are changing their variant names, there is then catch-all variants that show up as warnings for 'unused variables').

Byron (commented):
I don't perceive this as optimization, but that might just be an issue with the wording.

However, I don't think shortening lines is important enough to merge this PR, after all, the look doesn't change.

Thanks for your understanding.

PR #221, opened 2024-01-20
Fixes refresh with multiple input paths & adds tests

gosuwachu:


Byron:
That's so awesome, thanks a million for finding and fixing issues that I definitely should have caught earlier :D.

It's also great to have more test coverage than before, thanks for going the extra mile here!

Issue #68, opened 2020-10-25
Feature request: ARM64 release


mihaigalos:
Hi, awesome work. Can you also please release an ARM64 linux flavor?

mihaigalos:
(yes, I can compile it from source, but a binary on _each_ release would be awesome!)

Byron:
Thanks for the reminder, I disabled it at some point because it was failing for some reason. However, [now it is working again](https://github.com/Byron/dua-cli/releases/tag/TEST-0.0.1) and I have enabled it permanently.

The next regular release will have an arm version, and for completeness I have added the current arm build artifact to the most recent version.

_Note_ that I am not quite sure if it works on any linux since it's not a MUSL build, judging by the filename.

PR #229, opened 2024-01-25
feat: arm64 builds

chessmango:
Hello! Looking at https://github.com/Byron/dua-cli/issues/68 I see this was requested/fulfilled previously, though it seems releases have only ever been for `arm`, which is 32bit architecture specifically.

This PR adds aarch64/arm64 build support, using https://github.com/rust-cross/rust-musl-cross - I'm not overly familiar with rust and friends, so this may not be desired. Does seem to work as expected though (tested on the arm64 machine I'm wanting to use) - you can see/test results here: https://github.com/chessmango-forks/dua-cli/releases/tag/v2.28.3


Byron (approved):
Thanks a lot for tackling this, much appreciated :)!

PR #230, opened 2024-03-01
Ignore .vscode directory (same as .idea)

gosuwachu:


gosuwachu:
Will be ready after https://github.com/Byron/dua-cli/pull/231 lands on `main`.

Byron:
Thanks for the hint!

I actually want to go in the other direction, please see the commit message of 30f28a7426ea4093a0205e02674bd43943130b88 for details. Thanks for your understanding, but doing it like this will scale across all repositories right away without the need for changing each of their ignore files to match a local setup.

Doing so also enables [cool new features](https://github.com/Byron/gitoxide/discussions/1308) (*precious files*) :).

gosuwachu:
TIL, I have never used global `.gitignore`, but I have just set it up. Definitely useful for "precious files" :)

PR #233, opened 2024-03-09
Add scrollbar for main list

unixzii:
This PR adds a scrollbar for the main list to better indicate the current position of the viewport. See it in action:

https://github.com/Byron/dua-cli/assets/3109144/88cea506-fa29-4d80-95ad-be4b0e79740c

### DIFF
--- a/src/interactive/widgets/entries.rs
+++ b/src/interactive/widgets/entries.rs
@@ -11,10 +11,10 @@ use std::collections::HashSet;
 use std::time::SystemTime;
 use tui::{
     buffer::Buffer,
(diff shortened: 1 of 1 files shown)


Byron (approved):
That's a great addition, thanks so much!
It's also great to see it in action with the video :).

Since it seems quite straightforward to add thanks to `ratatui`, I wonder if you'd be interested to do the same for the 'mark' pane? If so, I'd hold back with a release for now.

unixzii:
Yep, I'd love to! I'll be filing another PR later today.

PR #235, opened 2024-04-05
Update gix-glob dependency to 0.16.2

blinxen:
I have run all tests locally and they pass. This should not introduce any problems (I think :D).


Byron:
Thanks! Just out of curiosity, Is this solving a particular problem?

blinxen:
I am preparing the new `helix` update in Fedora and need to update the huge `gitoxide` stack. This also means I have to update all packages that depend on older versions of `gitoxide` crates.

PR #246, opened 2024-05-26
Replace tui_react::Terminal with tui::Terminal

matta:
This is an experiment in converting a program that used the Terminal provided by tui-react into a program that uses the Terminal provided by ratatui. My hypothesis was that the two were not significantly different from eachother, and this seems to be true in this case.

I found that `Widget` trait used by ratatui's `Terminal` is sufficiently powerful to pass arbitrary mutable state to rendering code. A relatively simple `FunctionWidget` trampoline is used to achieve this.

The tui_react dependency is retained. The code still calls some tui_react utility functions and uses its List widget instead of the ratatui List widget. These utilities do not depend on the tui_react Terminal and could, in theory, be split out of tui-react into a separate crate.


Byron:
Thanks a lot for making this happen, and sorry for the hassle with clippy. It would have been fine to fix this here as well, but the respective PR #247 is on its way to be merged now.

Windows seems to have other issues due to `time` and `nightly` probably, maybe a regression in the compiler, so it's fair to just ignore that for now.

Byron (approved):
That's great work, thanks a lot! Will merge once CI is green.

Biswa96:
Would it be possible to release a new tag with this new dependencies? dua 2.29.0 fails to compile with rust 1.80.0 due to older time crate (< 0.3.36).

Byron:
Could you open an issue with details on how to reproduce the failure? I should then be able to take it from there. Thanks.

PR #250, opened 2024-07-16
doc: update version in install instructions

olastor:
Hi @Byron, thanks for this awesome cli-tool! I think the install tag might need to be updated here.


Byron (approved):
Thanks a lot, appreciated!

PR #271, opened 2024-12-13
style: simplify some statements for readability

### LINKED ISSUE #270: Some statements can be simplified for readability
For example, we can replace lines like

```rust
Some(format!("Deleted {} items...", entries_deleted));
```

with

```rust
Some(format!("Deleted {entries_deleted} items..."));
```

hamirmahal:
Closes #270

### DIFF
--- a/src/interactive/app/handlers.rs
+++ b/src/interactive/app/handlers.rs
@@ -244,8 +244,7 @@ impl AppState {
                         match self.delete_entry(entry_to_delete, tree_view) {
                             Ok(ed) => {
                                 entries_deleted += ed;
-                                self.message =
-                                    Some(format!("Deleted {entries_deleted} items..."));
+                                self.message = Some(format!("Deleted {entries_deleted} items..."));
                                 Ok(pane)
                             }
                             Err(c) => Err((pane, c)),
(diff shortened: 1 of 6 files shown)


hamirmahal (commented):
This shouldn't cause any user-facing changes.

The goal here is to improve code readability and maintainability.

hamirmahal:
`dua-cli` is an excellent command-line tool.

I like that development builds are really fast, especially relative to other Rust projects I've contributed to.

Would you want a check against build time regressions, to keep build times solid?

If you're interested, I can look into adding that to CI in a future pull request.

hamirmahal:
`dua-cli` is an excellent command-line tool.

I like that development builds are really fast, especially relative to other Rust projects I've contributed to.

Would you want a check against build time regressions, to keep build times solid?

I can look into adding that to CI in a future pull request if you're interested.

Byron (approved):
Thanks a lot, much appreciated!

> Would you want a check against build time regressions, to keep build times solid?

I think when it happens and you notice it, I would rather like to welcome a PR to fix the regression if it is fixable at all.

The reason I wouldn't want complexity to be added to CI for this is that I believe it would have to be timing based, and that will be flaky too easily.

Thanks for your understanding.

If you have more ideas on contributions you'd like to make, please feel free to do it with a PR right away - I really want to keep your overhead low and the extra issue here didn't seem necessary.

hamirmahal:
You're welcome! We could add some leeway, maybe 10% of the average build time of a few runs, to circumvent some flakiness, but it's completely up to you.

PR #274, opened 2025-01-20
Update petgraph to 0.7

musicinmybrain:
See https://github.com/petgraph/petgraph/blob/petgraph%40v0.7.1/RELEASES.rst. This is a SemVer-incompatible update only because `petgraph`’s exposed dependency `fixedbitset` was updated SemVer-incompatibly, but this doesn’t seem to require any code changes in dua-cli. I verified that `cargo test` still passes.


Byron (approved):
Thanks a lot!

Issue #126, opened 2022-05-12
Enhancement request: sort by name natually


wwhio:
It would be nice to have.

Sort by name:

```
chapter1.txt
chapter10.txt
chapter2.txt
chapter3p1.txt
chapter3p2.txt
chapter3p10.txt
chapter11.txt
chapter20.txt
```

Sort by name natually:

```
chapter1.txt
chapter10.txt
chapter11.txt
chapter2.txt       # we expect chapter 2 to be in front of chapter10
chapter20.txt
chapter3p1.txt
chapter3p10.txt
chapter3p2.txt
```

reference: https://blog.miigon.net/posts/how-to-sort-file-names-naturally/

PR #275, opened 2025-01-27
Feature: allow sorting by dirname / filename in interactive mode

### LINKED ISSUE #269: FR: Add "sort by name" in interactive mode
In interactive mode, we have
```
s => Toggle sort by size
m => Toggle sort by modified time
c => Toggle sort by entries
```

Sometimes, I'd like to sort by name.

Vaguely related: #83 and #126

joehasson:
Hi! This PR closes #269 by allowing users to toggle ascending / descending sort by name using the 'n' hotkey. A couple of notes on the design here, illustrated in the screen recording below.
- When the sort is first toggled, it is ascending by default, unlike the other sort modes, which default to descending (presumably users want to sort A-Z rather than Z-A)
- Entries are sorted first by type (directories before files) and lexicographically by name within each type. My thinking here is that users are likely sorting alphabetically so that they can navigate more easily. This is in line with how other command line tools like vim's file explorer perform sort by name.
- Names are sorted lexicographically rather than with a [natural sort](https://en.wikipedia.org/wiki/Natural_sort_order) as suggested in #126. Again this seems to be in line with how other tools sort by name, and is simpler to implement.
- The name column doesn't get highlighted in green once we choose to sort on it. I had a go at implementing this but think it may be more confusing than helpful given that names are highlighted to show which entries have / haven't been marked?

https://github.com/user-attachments/assets/b9fa83f0-9db9-4f44-a450-8996e91b151d

### DIFF
--- a/src/interactive/app/common.rs
+++ b/src/interactive/app/common.rs
@@ -15,6 +15,8 @@ pub enum SortMode {
     MTimeAscending,
     CountDescending,
     CountAscending,
+    NameDescending,
+    NameAscending,
 }
 
 impl SortMode {
@@ -42,6 +44,15 @@ impl SortMode {
             CountAscending => CountDescending,
             CountDescending => CountAscending,
             _ => CountDescending,
+        }
+    }
+
+    pub fn toggle_name(&mut self) {
+        use SortMode::*;
+        *self = match self {
+            NameAscending => NameDescending,
+            NameDescending => NameAscending,
+            _ => NameAscending,
         }
     }
 }
@@ -86,6 +97,15 @@ pub fn sorted_entries(
             .cmp(&r.entry_count)
             .then_with(|| l.name.cmp(&r.name))
     }
+    fn cmp_name(l: &EntryDataBundle, r: &EntryDataBundle) -> Ordering {
+        if l.is_dir && !r.is_dir {
+            Ordering::Less
+        } else if !l.is_dir && r.is_dir {
+            Ordering::Greater
+        } else {
+            l.name.cmp(&r.name)
+        }
+    }
     tree.neighbors_directed(node_idx, Direction::Outgoing)
         .filter_map(|idx| {
             tree.node_weight(idx).map(|entry| {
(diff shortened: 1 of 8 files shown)


Byron (approved):
This is great work, thanks a lot!

I will release it shortly!

Byron:
@EliahKagan I just wanted to share with you a finding I made just now. It's definitely a bug to be fixed in `gitoxide` at some point, maybe it's something that can be researched. My feeling is that it's something about how the dirwalk translates filesystem entries to precomposed unicode to look it up in the index (or its failure to do so).

```
dua-cli ( feat/interactive-mode-sort-by-name) [$?]
❯ gst
On branch feat/interactive-mode-sort-by-name
nothing to commit, working tree clean

dua-cli ( feat/interactive-mode-sort-by-name) [$?]
❯ gix status
 09:48:49 status done 111.0 files in 0.00s (39.3K files/s)
  ? tests/fixtures/right-to-left/نَاقَشَةُ سُبُلِ اِسْتِخْدَامِ اللُّغَةِ فِي النُّظُمِ الْقَائِمَةِ وَفِيم يَخُصَّ التَّطْبِيقَاتُ الْحاسُوبِيَّةُ،

dua-cli ( feat/interactive-mode-sort-by-name) [$?]
❯ gix clean
WOULD remove tests/fixtures/right-to-left/نَاقَشَةُ سُبُلِ اِسْتِخْدَامِ اللُّغَةِ فِي النُّظُمِ الْقَائِمَةِ وَفِيم يَخُصَّ التَّطْبِيقَاتُ الْحاسُوبِيَّةُ،

Skipped 1 directory - show with -d; Skipped 2 expendable entries - show with -x; Skipped 2 precious entries - show with -p
```

What seems to happen here is that the index lookup should tell `gix-dir` that the file is in the index and thus tracked, but the lookup fails.

Byron:
And here is the new release: https://github.com/Byron/dua-cli/releases/tag/v2.30.0 .

EliahKagan:
Did this start in gitoxide recently? Or is it just that it was recently discovered?

Byron:
I just discovered it, and would be thinking the issue is there from day one.

PR #284, opened 2025-05-06
Fix: Directories themselves do not contribute to size like they do in du

### LINKED ISSUE #262: Directories themselves do not contribute to size like they do in `du`
While experimenting with different distro images in Docker, I noticed that `du` and `dua` were a bit inconsistent (_and later that [`du` itself was inconsistent too with `--apparent-size`](https://github.com/Byron/dua-cli/discussions/121#discussioncomment-10541110) if the version of `du` was older_).

I discovered that `dua` reported `/usr/share/man` as only 20KiB, but `du` was showing about 1MB. This location had many directories nested and most empty. Other locations were like this as well (_I had been using `zypper --installroot /rootfs install glibc` for example to see how much space the package would weigh_), contributing several MB in weight invisible to `dua` for this small `glibc` package (2.7MB of 10.7MB total). This is due to each directory using a minimum 4096 bytes as the smallest block-size, with a new enough version of `du` the `--apparent-size` option will treat these as 0 bytes.

`dua` has the `-A` / `--apparent-size` flag too, but without it directory size seems to still be treated as 0 bytes. Perhaps that is useful for some users vs using `dua -A`? 🤷‍♂️ 

---

The linked discussion comment above suggested to raise an issue about this behaviour since it's technically a bug.

https://github.com/Byron/dua-cli/discussions/121#discussioncomment-10541462
> _It would probably be more correct if dua would also count empty directories as consumer of an inode, which typically clocks in at 4KiB._

---

It would be neat if `dua interactive` also supported toggling between `--apparent-size` btw, which as a bonus would make it available in the TUI help view?

joehasson:
Hi! This closes #262 where directory inode sizes are mistakenly being excluded from total disk usage calculations. This PR brings the behaviour of `dua` in line with that of `du` in both aggregate (`f44e19d`) and interactive (`ef1833a`) modes.

For example, given a directory tree structured as followed
```
/test_dir/
  ├── file1.txt (1KB)
  ├── empty_dir/
  └── populated_dir/
      └── file2.txt (2KB)
```
dua currently outputs
image
With these changes:
image

 Some notes on the changes made:
- In aggregate mode it suffices simply to remove the `!m.is_dir()` condition when traversing, as directory metadata can be handled just the same as non-directory file metadata
- In interactive mode, the changes are slightly more complex due to how entry size is recursively computed. We have existing logic for computing the size of a directory based solely on its contents (not counting the directory entry itself). In order to account for this entry size in the final total, I have added an `own_size` field to the `EntryData` struct, which stores the size just of the directory entry itself. Then, when updating our tree of `EntryData` nodes (in the `set_entry_info_or_panic` function), we simply add `own_size` to the total size of the directory's contents, which is computed just as before.
- No special cases  are needed to account for the --apparent-size flag. Following the lead of `du`, all directories are reported as consuming up a full block on disk regardless of whether reporting apparent size or disk usage.
-  Changing how entry counts are displayed: in interactive mode we are now displaying directory nodes in the entry_count as well as non-directory files. This makes it clearer that directory entry sizes are included in the calculations of the totals reported.

### DIFF
--- a/src/aggregate.rs
+++ b/src/aggregate.rs
@@ -52,8 +52,7 @@ pub fn aggregate(
                 Ok(entry) => {
                     let file_size = match entry.client_state {
                         Some(Ok(ref m))
-                            if !m.is_dir()
-                                && (walk_options.count_hard_links || inodes.add(m))
+                            if (walk_options.count_hard_links || inodes.add(m))
                                 && (walk_options.cross_filesystems
                                     || crossdev::is_same_device(device_id, m)) =>
                         {
--- a/src/interactive/app/tests/utils.rs
+++ b/src/interactive/app/tests/utils.rs
@@ -234,12 +234,12 @@ pub fn sample_01_tree() -> Tree {
     {
         let mut add_node = make_add_node(&mut tree);
         #[cfg(not(windows))]
-        let root_size = 1259070;
+        let root_size = 1275454;
         #[cfg(windows)]
         let root_size = 1259069;
-        let rn = add_node("", root_size, 10, None);
+        let rn = add_node("", root_size, 14, None);
         {
-            let sn = add_node(&fixture_str("sample-01"), root_size, 10, Some(rn));
+            let sn = add_node(&fixture_str("sample-01"), root_size, 14, Some(rn));
             {
                 add_node(".hidden.666", 666, 0, Some(sn));
                 add_node("a", 256, 0, Some(sn));
@@ -248,16 +248,26 @@ pub fn sample_01_tree() -> Tree {
                 add_node("c.lnk", 1, 0, Some(sn));
                 #[cfg(windows)]
                 add_node("c.lnk", 0, 0, Some(sn));
-                let dn = add_node("dir", 1258024, 5, Some(sn));
+
+                #[cfg(not(windows))]
+                let dn = add_node("dir", 1270312, 8, Some(sn));
+                #[cfg(windows)]
+                let dn = add_node("dir", 1258024, 8, Some(sn));
                 {
                     add_node("1000bytes", 1000, 0, Some(dn));
                     add_node("dir-a.1mb", 1_000_000, 0, Some(dn));
                     add_node("dir-a.kb", 1024, 0, Some(dn));
-                    let en = add_node("empty-dir", 0, 1, Some(dn));
+                    #[cfg(not(windows))]
+                    let en = add_node("empty-dir", 4096, 2, Some(dn));
+                    #[cfg(windows)]
+                    let en = add_node("empty-dir", 0, 2, Some(dn));
                     {
                         add_node(".gitkeep", 0, 0, Some(en));
                     }
-                    let sub = add_node("sub", 256_000, 1, Some(dn));
+                    #[cfg(not(windows))]
+                    let sub = add_node("sub", 260_096, 2, Some(dn));
+                    #[cfg(windows)]
+                    let sub = add_node("sub", 256_000, 2, Some(dn));
                     {
                         add_node("dir-sub-a.256kb", 256_000, 0, Some(sub));
                     }
@@ -274,8 +284,11 @@ pub fn sample_02_tree(use_native_separator: bool) -> (Tree, TreeIndex) {
     let root_index: TreeIndex;
     {
         let mut add_node = make_add_node(&mut tree);
+        #[cfg(not(windows))]
+        let root_size = 17924;
+        #[cfg(windows)]
         let root_size = 1540;
-        root_index = add_node("", root_size, 6, None);
+        root_index = add_node("", root_size, 10, None);
         {
             let sn = add_node(
                 format!(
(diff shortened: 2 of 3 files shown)


Byron (commented):
Thanks a lot for making this happen, it's much appreciated. I also love the PR description!

Do you think you could add a journey test that observes that directories now have a size?

Also, I wonder what adding another u128 to the structure does to its size. I am seeing the case where there are millions of files. Probably it doesn't matter as things like `name` will be responsible for most of the memory usage, but it would be good to have something like `assert_eq!(std::mem::size_of::<EntryData>(), X, "this shouldn't change unnoticed")` somewhere in the test-suite.

This is a dialogue and not a request, hence I am truly interested in what you think, and it would certainly be fine to merge as is otherwise.

Byron:
Edit: I just realized that CI is already failing because of it:

```
the size of this (96) should not exceed 80 as it affects overall memory consumption
```

Maybe we can give it a try and see what would happen if this was a boolean instead?

joehasson:
Thanks for taking a look. Yes, on reflection I agree it makes sense to be cautious about adding that extra field to the `EntryData` struct. I've had a bit of a think about the right way to proceed here.

### Alternative 1 - Still try to add directory inode size in `set_entry_info_or_panic`, but without storing the `own_size` field on `EntryData`

I see what you are suggesting regarding the boolean flag: if we know that an entry is a directory, we can add the directory inode's size to the `size` field whenever we update our tree of `EntryData` nodes, without having to also store the extra `own_size` field. Actually, we don't even need to check a boolean flag, since we only ever call `set_entry_info_or_panic` for directories anyway. The problem is that its not straightforward to compute the directory inode's size when we need it.

For example we might try using a constant as in the following
```
pub fn set_entry_info_or_panic(...) {
    ...
    // Doesn't work as directory inodes can vary in size
    node.size = size + DIR_INODE_SIZE_CONSTANT;
   ...
}
```
but the directory inode size may vary depending on the number of children it has. I also thought about grabbing the inode size from the file metadata here e.g. like this
```
pub fn set_entry_info_or_panic(...) {
   ...
   let node_own_size = std::fs::metadata(&node.name)
        .map(|m| m.len())
        .expect("This panics because node.name is the immediate directory name not a path")
    as u128;
   
    node.size = size + node_own_size;
    ...
}
```
but we need a proper path here, not just the directory's name. Presumably we don't want to start storing that on the `EntryData` structure for the same reasons we don't want to start storing `own_size`.  Also, computing the directory entry size like this seems redundant, since we already process the directory metadata once the `TraversalEvent` for that directory arrives.
### Alternative 2 - Track directory inode sizes during recursive traversal of the filesystem
The original PR changes and Alternative 1 both leave untouched the algorithm which recursively computes the size of the directory's contents. This approach would add three fields to the `BackgroundTraversal` struct to augment the existing recursive traversal:
- `parent_node_size_per_depth_level`,  is a stack of directory inode sizes for ancestor nodes along the current branch
- `parent_node_size`, analogous to `parent_node_idx`,  is the size of the parent node of the entry we are walking
- `previous_node_size`, analogous to `previous_node_idx`, used to updated `parent_node_size` when we descend a level in the recursive traversal

We then are able to pass the `parent_node_size` to `set_entry_info_or_panic` alongside the `current_directory_at_depth`, to correctly calculate the size of the tree entry.

Using this approach, the extra cost we pay for storing the directory sizes is linear in the height of the file tree, rather than paying a cost for every single file in the file-system. This seems more reasonable to me.

 I have created another branch using this approach [here](https://github.com/Byron/dua-cli/commit/a81b892ee289da2e2e9cf63caebf44a2e4f3b029). Happy to take your thoughts on any of this - if you agree that alternative 2 looks good then I will finalise this version, update the PR and go about adding that journey test you mentioned. Cheers!

Byron:
That's great, if you prefer option 2 I will happily go with that. There might be an option 3, which is to real-world test the impact of the added 16 bytes. As option 2 will add complexity, I think we should benchmark the memory usage first, and then see what by what percentage the consumption rises.

Also, please feel free to push these changes here right away, preferably in their own commit, then it's more straightforward to keep this endeavour in one place.
Thanks again.

Byron:
Thank you - once the numbers for Windows are adjusted, could you please squash the commits down into one?
Then there should be nothing in the way of merging it. 
Thanks again.

joehasson:
Cool thank you. To me these updated tests capture the fact that directories now have size, so no new journey test is needed. Do you agree?
As for the memory usage issue, I did some memory profiling and found that the peak heap utilisation for the original implementation was ~15% higher than the latest release of dua, compared to only a 2-3% increase for the improved implementation, so I think the slight added complexity is worth it. This was running against my own $HOME directory of around 560k files - I can share the script I used to do this profiling if you would like to take a look.
I see that the tests are now passing on linux but still failing on windows - I'll take a look as to why.

joehasson:
Fixed and squashed - hopefully good to go now. Cheers.

Byron (approved):
Fantastic work, thank you!

Byron:
And here is the new release: https://github.com/Byron/dua-cli/releases/tag/v2.30.1 .

PR #287, opened 2025-06-08
Toggle UI split

gabrielfalcao:
Hi @Byron,

I added a feature to toggle the UI direction because that's useful for long file names in small monitors.

Thanks for making dua-cli!

### DIFF
--- a/src/interactive/app/eventloop.rs
+++ b/src/interactive/app/eventloop.rs
@@ -220,6 +220,11 @@ impl AppState {
         EntryCheck::new(self.scan.is_some(), self.allow_entry_check)
     }
 
+    fn toggle_ui_split(
+        &mut self,
+    ) {
+        self.ui_split = !self.ui_split;
+    }
     fn process_terminal_event<B>(
         &mut self,
         window: &mut MainWindow,
@@ -272,6 +277,9 @@ impl AppState {
                 if let Some(result) = self.handle_quit(&mut tree_view, window) {
                     return Ok(Some(result?));
                 }
+            }
+            Char('u') => {
+                self.toggle_ui_split();
             }
             _ => {
                 handled = false;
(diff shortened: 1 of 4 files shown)


gabrielfalcao:
@Byron I made the changes but am having a hard time understanding why the CI is failing. Would you kindly shed some light there for me?

src/interactive/widgets/entries.rs
```diff
@@ -192,7 +192,7 @@ fn title(
 
 fn draw_bottom_right_help(bound: Rect, buf: &mut Buffer) {
     let bound = line_bound(bound, bound.height.saturating_sub(1) as usize);
-    let help_text = " mark-move = d | mark-toggle = space | toggle-all = a ";
```
Byron, line 195: The trailing space is needed for… spacing :).

src/interactive/widgets/help.rs
```diff
@@ -179,6 +179,7 @@ impl HelpPane {
                 );
                 hotkey("<Space>", "Toggle the currently selected entry.", None);
                 hotkey("a", "Toggle all entries.", None);
+                hotkey("u", "Toggle UI split Vertical/Horizontal.", None);
```
Byron, line 182: This should be with `Pane Control`.

src/interactive/app/eventloop.rs
```diff
@@ -273,6 +278,9 @@ impl AppState {
                     return Ok(Some(result?));
                 }
             }
+            Char('u') => {
```
Byron, line 281: This letter is already taken.

Byron (changes requested):
Thanks! This time I tried it and it works as advertised, it's definitely good to have such 'pane-rotation'.

It looks like CI is generally broken now for some reason and I wonder if @joehasson has more intuition than me.

Disregarding CI, I have a few more comments that would need addressing.
Thanks again.

src/interactive/app/eventloop.rs
Byron: I think you will have to try out what feels right on an english keyboard.

joehasson:
@Byron Regarding CI, I actually don't think it is broken generally: I've been able to reproduce the failing tests on linux locally by bisecting to the changes introduced in commit 19e1e17 of this PR (I don't have a windows machine to try on but I imagine the culprit is the same). Its not clear to me why these changes would cause the tests to fail though.

PR #290, opened 2025-08-01
Prompt user before quitting

unixzii:
Dua will quit directly when pressing esc if the focus is on the main pane. This may cause users to accidentally quit the app when they only want to dismiss some other panels. It's especially frustrating if the scan took a long time.

This PR adds a prompt that shows before quitting the app. When you pressed esc or q, the status bar will show the prompt first. To really quit, you need to press esc or q again. You can also cancel the quit operation by pressing any key else. Meanwhile, ctrl-c still quits the app directly since it's a combination key.

There is a screencast to show how it works in action:

https://github.com/user-attachments/assets/38e70df9-fd58-4a04-915e-832d7edb5154

### DIFF
--- a/src/interactive/app/eventloop.rs
+++ b/src/interactive/app/eventloop.rs
@@ -249,13 +249,21 @@ impl AppState {
 
         let glob_focussed = self.focussed == Glob;
         let mut tree_view = self.tree_view(traversal);
-        let mut handled = true;
-        match key.code {
-            Esc => {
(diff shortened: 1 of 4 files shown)


Byron (approved):
That's a great feature, thanks a lot!

I will create a new release shortly.

Byron:
And here it is: https://github.com/Byron/dua-cli/releases/tag/v2.31.0

unixzii:
Thanks!

PR #321, opened 2026-02-22
Add command to install without cloning

pcgeek86:
Add command to README, to install without cloning


Byron:
Sorry, I don't want to recommend installing unreleased versions. Thanks for your understanding.

PR #326, opened 2026-04-19
fix: respect CARGO_HOME in install.sh

### LINKED ISSUE #279: Installation should take CARGO_HOME into account for default location
I have configured `CARGO_HOME` to be `$HOME/.cache-cargo` and installed `dua` following the installation instructions. The target location where `dua` ended was a surprise for me.

Following the [pattern of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) I would have expected to end up in `$CARGO_HOME/bin`.

```
➜ echo $CARGO_HOME
/home/max/.cache/cargo
➜ curl -LSfs https://raw.githubusercontent.com/Byron/dua-cli/master/ci/install.sh | \
    sh -s -- --git Byron/dua-cli --target x86_64-unknown-linux-musl --crate dua --tag v2.29.0
install.sh: GitHub repository: https://github.com/Byron/dua-cli
install.sh: Crate: dua
install.sh: Tag: v2.29.0
install.sh: Target: x86_64-unknown-linux-musl
install.sh: Installing to: /home/max/.cargo/bin
install.sh: Downloading: https://github.com/Byron/dua-cli/releases/download/v2.29.0/dua-v2.29.0-x86_64-unknown-linux-musl.tar.gz
'/tmp/tmp.GSDRgpom8t/./dua-v2.29.0-x86_64-unknown-linux-musl/dua' -> '/home/max/.cargo/bin/dua'
```

Suggestion: if `CARGO_HOME` is defined, use it a default location.

lawrence3699:
Fixes #279.

Before this change, omitting `--to` always defaulted the installer to `$HOME/.cargo/bin`, even when `CARGO_HOME` was set. That made the documented installer path disagree with the user's configured Cargo home.

This updates the default destination to use `$CARGO_HOME/bin` when `CARGO_HOME` is set, and keeps the existing `$HOME/.cargo/bin` fallback otherwise. The journey test coverage now exercises both branches with fixed `HOME`/`CARGO_HOME` values.

Validation:
- manual repro with fake `HOME`/`CARGO_HOME` before and after the change
- `./tests/stateless-journey.sh ./target/debug/dua`
- `cargo fmt --all -- --check`
- `cargo clippy -- -D warnings`
- `cargo test --all` still failed locally in two unrelated interactive size assertions on this macOS environment

### DIFF
--- a/ci/install.sh
+++ b/ci/install.sh
@@ -16,7 +16,7 @@ Options:
     --crate NAME    Name of the crate to install (default <repository name>)
     --tag TAG       Tag (version) of the crate to install (default <latest release>)
     --target TARGET Install the release compiled for $TARGET (default <`rustc` host>)
-    --to LOCATION   Where to install the binary (default ~/.cargo/bin)
+    --to LOCATION   Where to install the binary (default $CARGO_HOME/bin or ~/.cargo/bin)
 EOF
 }
 
@@ -130,7 +130,7 @@ fi
 say_err "Target: $target"
 
 if [ -z $dest ]; then
-    dest="$HOME/.cargo/bin"
+    dest="${CARGO_HOME:-$HOME/.cargo}/bin"
 fi
 
 say_err "Installing to: $dest"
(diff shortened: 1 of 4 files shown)


Byron (changes requested):
Thanks a lot!

Let's keep only the one change to `install.sh`, then I can merge.

PR #335, opened 2026-06-06
Add arm64 Windows support

knightmare2600:
github runner updated to also build a native arm64 binary for windows

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,9 +11,7 @@ permissions:
 
 jobs:
 
-  # =========================
-  # CREATE RELEASE (ONCE ONLY)
-  # =========================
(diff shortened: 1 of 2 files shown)


Byron:
Thanks, but I'm afraid this can't be done as completely right. I can't review this, nor can I test it.

But what killed it is that it seems to just quietly remove the auto-release-on-tag feature, which I need.

knightmare2600:
Hi,

Sorry, I'm new to github PRs and so on with workflows attached.

If I was to re-add auto-tag on release and provide some evidence, e.g. it running inside a windows ARM device, would you consider re-opening..?

Thanks

Byron:
> If I was to re-add auto-tag on release and provide some evidence, e.g. it running inside a windows ARM device, would you consider re-opening..?

No, it can be a minimal PR only.

PR #341, opened 2026-06-23
Update release.yml

knightmare2600:
Hi,

We spoke a few weeks ago about an arm64 build. I *think* this is the minimal changes without breaking the workflows and tags you requested.

I have an M1 Macbook (arm CPU) and it runs fusion, so I'm happy to run the binary or provide any evidence if required.

Thanks,!


Byron (approved):
That's it in principle, and I think it should work after my fixup. If not, I can revert it, so worth a try.

PR #345, opened 2026-07-07
Add release for linux-riscv64

luhenry:
I verified it works at https://github.com/luhenry/dua-cli/actions/runs/28885540933


Byron (approved):
Thanks! Let's try it.

PR #349, opened 2026-07-15
Suggestion of features + fixes

gabrielfalcao:
- Interactive Mark: switch between split horizontal and vertial via hotkey (uppercase) "U"
- Standard Ouput after quitting Interactive Mark: print absolute paths as opposed to relative paths.
- Aggregate Output: only print colored output when stdout supports just so.

This squashed commit has been bisected when the upstream head is fa5cd4540b762d63acf6c4ff4de78b6d41f2ebcb and the last common ancestor is 63b129b1addbac7f4b238529875d88062ab68dfb

The following commits which might or might not remain in the reflog are:

commit 876a54fe9a1ec06044c76ace9a45d30e24250b2c
commit 529e0ea67974bc8a6db577aed13be0abc102db1b
commit 169eade96752b9acce01ee62d7acccc7d60cf327
commit 83d987dcce4d995849fb201043d146026465be81
commit a3a682d46d527ba1f28427c21f3732238d29d62f
commit f8deb24c01e88a49a2122073afbfb650ce30ed20
commit 140bf8df6f2467ce0bb5d0e6cefc1723ae554841
commit b80ad4217744305a42333763abf382bd51d49c59
commit 19e1e176c1deebecec1500dc2445d664e2db395a

Conflicts:
	Cargo.lock
	Cargo.toml
	src/interactive/app/state.rs
	src/interactive/widgets/entries.rs
	src/interactive/widgets/help.rs
	src/interactive/widgets/main.rs
	src/main.rs


Byron:
I don't know what happened here, but this is too sloppy and strange.

gabrielfalcao:
> I don't know what happened here, but this is too sloppy and strange.

Just for the record, no A.I. involved, any sloppyness here is my own 🥂 
No harm no foul, thanks for creating this amazing open source tool. I use it in a daily basis and being able to build my own fork is a factor of two things: This project being open source AND your source code being very easy to navigate, understand and modify.

You l33t, kudos!

Byron:
Hehe, thanks! I wasn't even aware that calling something "sloppy" might be insulting, so my apologies for this. Nonetheless, if you want to upstream changes, it's certainly better to keep them small and well bounded.

> Aggregate Output: only print colored output when stdout supports just so.

This seems like a bugfix that should be upstreamed.

> Standard Ouput after quitting Interactive Mark: print absolute paths as opposed to relative paths.

I think this would make it more universally useful, and easier to use in consuming programs so let's have it, in the same PR is fine, but a separate commit.

> Interactive Mark: switch between split horizontal and vertial via hotkey (uppercase) "U"

Let's see how it looks in a video - this could be the last PR. Generally I like here that ultimately it removes code, which makes this PR desirable in principle. With a little more care we should be able to get everything merged.

PR #350, opened 2026-07-20
fix: keep interactive cursor visible under NO_COLOR

### LINKED ISSUE #238: Cursor line not visible with NO_COLOR
when I have `NO_COLOR=true` in the environment, `dua i` won't display the cursor line:

It also seems to appear when I press `C` to display file count:

But even then, sometimes it displays the cursor line when moving up, but disappears when moving down.  

It works fine if I:

```
unset NO_COLOR
dua i
```

Solaris-star:
Under `NO_COLOR`, the selected entry in `dua i` was invisible because REVERSED was reset by empty SGR color sequences from crossterm. Prefer UNDERLINED (and keep BOLD when focused) when `NO_COLOR` is set.

Fixes #238.

### DIFF
--- a/src/interactive/widgets/entries.rs
+++ b/src/interactive/widgets/entries.rs
@@ -476,7 +476,15 @@ fn draw_top_right_help(area: Rect, title: &str, buf: &mut Buffer) -> Rect {
 fn style(is_selected: bool, is_focussed: bool) -> Style {
     let mut style = Style::default();
     if is_selected {
-        style.add_modifier.insert(Modifier::REVERSED);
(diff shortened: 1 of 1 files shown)


Byron:
For me this doesn't do anything, and `NO_COLOR=1` + this PR is the same as without it.

Screenshot 2026-07-20 at 09 31 06

The bar isn't visible, because this fix suffers from the same issue as before: the style is reset immediately.
Let's see if there is a simple solution to this.

Byron:
Superseded by #351 .

PR #352, opened 2026-07-23
fix: interactive error handling and footer status polish

l0rush1:
- Non-zero exit + printed error on interactive failure
- Guard footer entries/s when elapsed is 0
- Show open failures in the footer
- Label cleanup vs gitignored counts correctly


Byron (approved):
Thanks a lot!

The changelog entries for the commits won't be easy to understand, but will probably be enough as a hint.

PR #353, opened 2026-07-28
refactor(delete): use jwalk for parallel, symlink-safe deletion

### LINKED ISSUE #43: parallel deletion (without following symlinks) using jwalk
Previously jwalk would follow symlinks, so we could not use it and instead implemented our own traversal.

Now with jwalk 0.5, this is supported and we can use our standard jwalker to do the job.

https://github.com/Byron/dua-cli/blob/543f7f3948c26250a8fc6ebf79a49f3ddfa3cb63/src/interactive/app/handlers.rs#L305-L340

The function above would have to change to use jwalk. It should be straightforward to obtain a walker with this function:

https://github.com/Byron/dua-cli/blob/0d6116eea1e741bc8bc1fc6d04536c8242c5aa42/src/common.rs#L162-L184

Solaris-star:
## Summary

Fixes #43

`delete_directory_recursively` used a hand-rolled stack traversal because, when this issue was filed, jwalk followed symlinks and could not be used safely for deletion. jwalk has supported `follow_links(false)` for a long time now, so this switches the deletion walk over to the same `jwalk::WalkDir` the rest of dua already uses for scanning — giving parallel traversal on multi-core machines.

### Behaviour preserved

- **Symlinks are not followed** — a symlink is removed with `remove_file` on the link itself; its target is left untouched.
- **Directories are removed deepest-first** (sorted by descending depth) so each `remove_dir` sees an empty directory.
- **Error counting and byte accounting are unchanged** — traversal errors increment `stats.errors`, and per-entry bytes are still taken from `symlink_metadata().len()`.

### Tests

Added `delete_directory_recursively_tests` covering:

- [x] single file removal
- [x] nested directory tree (files + dirs, deepest-first)
- [x] symlink removed without deleting its target (unix)
- [x] missing path reports an error

```
cargo test delete_directory_recursively_tests
running 4 tests
test ...::reports_an_error_for_a_missing_path ... ok
test ...::removes_a_single_file ... ok
test ...::removes_symlink_without_following_it ... ok
test ...::removes_a_nested_tree ... ok
test result: ok. 4 passed; 0 failed
```

`cargo fmt --check` and `cargo build` are clean.

> Note: two pre-existing tests in `interactive::app::tests::unit` (`it_can_handle_ending_traversal_*`) fail on `main` in my environment before this change — unrelated to deletion.

### DIFF
--- a/src/interactive/app/handlers.rs
+++ b/src/interactive/app/handlers.rs
@@ -6,10 +6,12 @@ use crate::interactive::{
 use crossterm::event::KeyEvent;
 use dua::Config;
 use dua::traverse::TreeIndex;
+use jwalk::rayon::prelude::*;
 use std::{
     collections::BTreeSet,
     fs, io,
     path::PathBuf,
+    sync::Arc,
     time::{Duration, Instant},
 };
 use tui::{Terminal, backend::Backend};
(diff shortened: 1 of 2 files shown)


Byron:
Fantastic, thanks a lot! 

I will create a release shortly.

PR #366, opened 2026-08-12
fix: sanitize control characters in marked path output

### LINKED ISSUE #365: Marked path printed on quit isn't stripped of terminal escape sequences, allowing injection via a crafted file name
Affected versions: confirmed on the tip of `master`, commit b6e7caf (2026-08-05), released version 2.41.1

## Summary

dua-cli's interactive TUI is built on ratatui, which protects the paths it renders on screen. However, marking a file for deletion (space bar) and then quitting prints that file's path directly to the terminal after the TUI has already released terminal control, bypassing ratatui's protective rendering entirely. A scanned file's name has no character restrictions, so marking a crafted file and quitting writes any escape sequence embedded in its name to the terminal, where it's interpreted live.

## Details

`src/main.rs`, in `main`, reached after `drop(terminal)` / `drop(terminal_guard)` have already released ratatui's screen control:
```rust
let exit_code = match res {
    Ok((walk_result, paths)) => {
        if let Some(paths) = paths {
            for path in paths {
                println!("{}", path.display());
            }
        }
        walk_result.to_exit_code()
    }
    ...
```
`paths` here comes from `app.window.mark.take().map(|pane| pane.into_paths())`, the set of entries the user marked with the space bar during the session. No filtering of control characters happens before the print.

## POC

(available upon request)

## Impact

dua-cli writes a marked entry's path to the terminal exactly as read from the filesystem, with no stripping of control characters or escape sequences, after the TUI has exited. Since running `dua i` on a directory and marking a file before quitting is ordinary use of the tool's own core feature, a crafted file name causes any embedded escape sequence to be written to the terminal and interpreted live. I demonstrated this with a benign OSC title-set marker rather than a more aggressive payload, since confirming the sequence reaches output unmodified is sufficient to establish the defect; the specific range of what an injected OSC/CSI sequence can achieve depends on the victim's terminal emulator and wasn't enumerated further.

I have a fix ready and will follow up with a PR.

carfeii:
Fixes #365.

dua-cli's interactive TUI is built on ratatui, which protects the paths it renders on screen. But marking a file for deletion and then quitting prints that file's path directly to the terminal after the TUI has already released terminal control, bypassing ratatui's protective rendering entirely. A scanned file's name has no character restrictions, so a crafted file name can inject terminal escape sequences into the printed path.

Adds sanitize_for_display (src/main.rs), replacing control characters with the Unicode replacement character, applied at the sole print site for marked paths.

Verified against a file named using a raw OSC title-set escape sequence, marking it and quitting, output captured via tmux's raw pty pipe and inspected as a hex dump. Confirmed no regression with a normal file name. Full test suite (72 tests) passes.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -59,6 +59,17 @@ impl Drop for InteractiveTerminalGuard {
     }
 }
 
+/// Replaces control characters (which can include terminal escape
+/// sequences) with the Unicode replacement character. A marked entry's
+/// path has no character restrictions, and is printed to the real
+/// terminal after the TUI has already released terminal control,
+/// bypassing ratatui's own protective rendering entirely.
+fn sanitize_for_display(s: &str) -> String {
+    s.chars()
+        .map(|c| if c.is_control() { '\u{FFFD}' } else { c })
+        .collect()
+}
+
 fn main() -> Result<()> {
     #[cfg(feature = "tui-crossplatform")]
     use options::Command::Interactive;
@@ -176,7 +187,7 @@ fn main() -> Result<()> {
                 Ok((walk_result, paths)) => {
                     if let Some(paths) = paths {
                         for path in paths {
-                            println!("{}", path.display());
+                            println!("{}", sanitize_for_display(&path.display().to_string()));
                         }
                     }
                     walk_result.to_exit_code()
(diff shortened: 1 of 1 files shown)


Byron:
@codex review

Byron:
Thanks, good catch!
The interactive display is safe via ratatui, and this is the spot where it's still lacking.

PR #367, opened 2026-08-14
Avoid per-entry macOS metadata queries

tamird:
Directory iteration on macOS fetched metadata separately for every
entry. Expanded roots then discarded that metadata, and output
coloring queried each path again.

Collect metadata with getattrlistbulk and preserve native entries
through aggregate and interactive traversal. Reuse each root file
type when formatting output instead of querying its path again.

A representative aggregate scan drops from 532 ms to 108 ms,
roughly 4.9 times faster.

Native metadata replaces std::fs::Metadata on macOS, requiring
dua-core 3.0. Preserve directory permission errors, modification
times, device boundaries, symlinks, hard links, and resource forks.

This code was written almost entirely by LLM with my guidance! I will not be
offended if you reject it, rewrite it from scratch, or ask for significant
changes!

### DIFF
--- a/crates/dua-lib/src/macos/mod.rs
+++ b/crates/dua-lib/src/macos/mod.rs
@@ -23,7 +23,7 @@ const DIRECTORY_BUFFER_BYTES: usize = 64 * 1024;
 
 /// A macOS filesystem entry produced from native directory metadata.
 pub struct Entry {
-    /// Distance from the walk root: `0` for the root, `1` for its children, and so on.
+    /// Distance from the walk root: `0` for the root, `1` and highger for its children.
     pub depth: usize,
     /// File name relative to `parent_path`.
     pub file_name: OsString,
@@ -175,6 +175,7 @@ impl Metadata {
 
 pub(crate) struct ReadDir {
     directory: OwnedFd,
+    fallback: Option<fs::ReadDir>,
     buffer: Box<AlignedBuffer<DIRECTORY_BUFFER_BYTES>>,
     offset: usize,
     remaining: usize,
@@ -193,6 +194,7 @@ impl ReadDir {
             .into();
         Ok(Self {
             directory,
+            fallback: None,
             buffer: Box::new(AlignedBuffer::new()),
             offset: 0,
             remaining: 0,
@@ -203,6 +205,11 @@ impl ReadDir {
         })
     }
 
+    /// Refill the bulk-record buffer or activate ordinary directory iteration when the filesystem
+    /// does not support bulk enumeration.
+    ///
+    /// Returns `true` when iteration can continue, either because bulk records are available or
+    /// because `self.fallback` was initialized. Returns `false` when the directory is exhausted.
     fn refill(&mut self) -> io::Result<bool> {
         loop {
             let mut attributes = requested_attributes(self.listing_error.is_some());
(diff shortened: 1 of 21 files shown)


Byron:
Thanks so much, I wasn't aware that macOS also has a special API just like Windows.

On my standard test tree, it's a reasonable 1.34x improvement, nothing to sneeze at.
Also note the massively reduced user and system time.

## Benchmarks

### implicit `-t8`
```
❯ hyperfine 'dua ../..' './target/release/dua ../..'
Benchmark 1: dua ../..
  Time (mean ± σ):     10.693 s ±  1.846 s    [User: 3.345 s, System: 49.738 s]
  Range (min … max):    9.911 s … 15.927 s    10 runs

Benchmark 2: ./target/release/dua ../..
  Time (mean ± σ):      7.984 s ±  0.188 s    [User: 1.551 s, System: 26.065 s]
  Range (min … max):    7.782 s …  8.390 s    10 runs

Summary
  ./target/release/dua ../.. ran
    1.34 ± 0.23 times faster than dua ../..
```

### `-t4`

```
dua-cli.macos-native-traversal on  HEAD (2ebd21d) +711 -126 [$!?] is 📦 v2.41.1 took 3m6s
❯ hyperfine 'dua -t4 ../..' './target/release/dua -t4 ../..'
Benchmark 1: dua -t4 ../..
  Time (mean ± σ):     14.777 s ±  0.101 s    [User: 2.329 s, System: 29.944 s]
  Range (min … max):   14.573 s … 14.953 s    10 runs

Benchmark 2: ./target/release/dua -t4 ../..
  Time (mean ± σ):     10.800 s ±  0.745 s    [User: 1.200 s, System: 17.696 s]
  Range (min … max):   10.383 s … 12.877 s    10 runs

Summary
  ./target/release/dua -t4 ../.. ran
    1.37 ± 0.09 times faster than dua -t4 ../..
```

### `-t16`

Contention is very similar for both, but the new version is still faster.

```
❯ hyperfine 'dua -t16 ../..' './target/release/dua -t16 ../..'
Benchmark 1: dua -t16 ../..
  Time (mean ± σ):     12.042 s ±  0.249 s    [User: 4.577 s, System: 135.404 s]
  Range (min … max):   11.736 s … 12.504 s    10 runs

Benchmark 2: ./target/release/dua -t16 ../..
  Time (mean ± σ):     11.833 s ±  0.585 s    [User: 2.168 s, System: 110.774 s]
  Range (min … max):   10.460 s … 12.551 s    10 runs

Summary
  ./target/release/dua -t16 ../.. ran
    1.02 ± 0.05 times faster than dua -t16 ../..
```

### `-t1`

Interesting - the 30% performance improvement translates to each thread.

```
❯ hyperfine 'dua -t1 ../..' './target/release/dua -t1 ../..'
Benchmark 1: dua -t1 ../..
  Time (mean ± σ):     47.790 s ±  0.936 s    [User: 2.128 s, System: 24.485 s]
  Range (min … max):   46.752 s … 49.796 s    10 runs

Benchmark 2: ./target/release/dua -t1 ../..
  Time (mean ± σ):     36.669 s ±  0.780 s    [User: 1.215 s, System: 15.129 s]
  Range (min … max):   35.702 s … 37.952 s    10 runs

Summary
  ./target/release/dua -t1 ../.. ran
    1.30 ± 0.04 times faster than dua -t1 ../..
```

Byron:
During the review I removed InputPath as it makes no difference in my setup.
If you have `hyperfine` benchmarks to show the difference in your case, a follow-up PR could bring it back.

tamird:
Thank you for carrying this the rest of the way!

crates/dua-lib/src/macos/mod.rs
```diff
@@ -23,7 +23,7 @@ const DIRECTORY_BUFFER_BYTES: usize = 64 * 1024;
 
 /// A macOS filesystem entry produced from native directory metadata.
 pub struct Entry {
-    /// Distance from the walk root: `0` for the root, `1` for its children, and so on.
+    /// Distance from the walk root: `0` for the root, `1` and highger for its children.
```
tamird, line 26: Typo

Byron: That's the one bit I wrote myself :D

tamird: We're obsolete D:

tamird:
The missing case is a wide, shallow directory rather than a deep tree. dua treats each immediate child as an aggregation root, so removing the prepared metadata adds a per-child lookup for symlink filtering and another for root initialization; --stay-on-filesystem adds further device lookups.

On an existing Cargo build directory with 47,598 immediate files, current main takes 460.9 ms versus 99.7 ms with prepared roots (4.62x faster). With --stay-on-filesystem, it is 996.3 ms versus 94.4 ms (10.56x faster). A deep-tree control remains effectively unchanged: 156.5 ms versus 157.7 ms.

The focused follow-up is #369. It reuses bulk metadata only for aggregation and does not restore the InputPath wrapper or interactive plumbing.

PR #369, opened 2026-08-14
Reuse bulk metadata for macOS aggregation roots

tamird:
When dua expands a directory into one aggregation root per immediate child, macOS currently looks up each child's metadata once for symlink filtering and again when initializing the traversal root. --stay-on-filesystem adds separate filesystem identity lookups. That overhead is negligible on a deep tree with few immediate children, but dominates wide, flat directories.

Reuse the metadata and device identity already returned by getattrlistbulk for aggregation roots. Keep existing path-based APIs and interactive traversal unchanged; pass the existing dua_core::Entry directly instead of restoring the InputPath wrapper removed in #367 [1].

On an existing Cargo build directory containing 47,598 immediate files:
```
Workload                  Current main   This change   Improvement
Regular aggregation         460.9 ms       99.7 ms         4.62x
--stay-on-filesystem        996.3 ms       94.4 ms        10.56x
Deep-tree control           156.5 ms      157.7 ms     unchanged
```
The additional traversal entry points are backward compatible and require dua-core 3.1.0. Publish dua-core 3.1.0 before packaging the next dua-cli release.

[1]: https://github.com/Byron/dua-cli/pull/367#issuecomment-5293310504

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -710,7 +710,7 @@ dependencies = [
 
 [[package]]
 name = "dua-core"
-version = "3.0.0"
+version = "3.1.0"
 dependencies = [
  "crossbeam",
  "libc",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,7 @@ tui-crossplatform = [
 trash-move = ["trash"]
 
 [dependencies]
-dua-core = { path = "crates/dua-lib", version = "3.0.0" }
+dua-core = { path = "crates/dua-lib", version = "3.1.0" }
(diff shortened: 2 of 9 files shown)


tamird:
Standard caveat here: code written by LLM, please take/discard/reject/rewrite whatever you like!

Byron:
Great, thanks a lot!

Performance is still the same in my test-case, which is just to show it didn't get worse in cases where the optimization doesn't matter.

I have also added this capability to Windows, while at it.

PR #370, opened 2026-08-15
Apply --ignore-dirs to paths dua expanded into roots itself

VXNCXNX:
## What's broken

`--ignore-dirs` has no effect when you pass a single directory, which is the common invocation. The help text says these paths "will only be ignored if they are eventually reached as part of the traversal", and the ignored directory is reached, but it is still measured.

## Repro

```
$ dua aggregate -i /tmp/duatest/root1/ignoreme /tmp/duatest/root1
   12.00 KiB keep
   56.00 KiB ignoreme
   68.00 KiB total
```

Passing two roots instead works, which is what pointed at the cause.

## The fix

Given zero or one directory, `extract_paths_maybe_set_cwd` chdirs into it and expands its children into independent traversal roots. Roots are deliberately exempt from the `-i` check in `common.rs`, since an explicitly passed path should never be ignored. Children promoted to roots by that expansion inherit the exemption they were never meant to have.

The filter now applies to expanded paths only, alongside the existing `ignore_patterns` filter in the same place, and short-circuits when `ignore_dirs` is empty so the common case costs nothing. Explicitly passed paths keep the documented behaviour: `dua aggregate -i /x /x` still measures `/x`.

This also affects `dua i`, which goes through the same function.

## Verification

Two cases in `tests/stateless-journey.sh` with snapshots: a directory below the given path is dropped, and the same directory named explicitly is still measured. Both snapshots fail without the change, the first one showing `dir` back in the output. `cargo test --all` and the journey suite pass.

I put these in the journey suite rather than as unit tests on purpose. The function changes the process' current directory, and a unit test doing that raced `journeys_readonly::once_waits_for_replayed_refresh_to_finish` under the parallel runner, failing about one run in five.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -411,6 +411,10 @@ fn extract_paths_maybe_set_cwd(
     walk_options: &dua::WalkOptions,
 ) -> Result<Vec<PathBuf>, io::Error> {
     let cross_filesystems = walk_options.cross_filesystems;
+    // Paths were explicitly passed by the user on the command-line; per `--ignore-dirs`'s
+    // documented behavior, these are never subject to `-i`/`--ignore-dirs` filtering, only
+    // paths that we ourselves expand into roots below are.
+    let paths_were_expanded = paths.is_empty() || (paths.len() == 1 && paths[0].is_dir());
     if paths.len() == 1 && paths[0].is_dir() {
         std::env::set_current_dir(&paths[0])?;
         paths.clear();
@@ -439,6 +443,15 @@ fn extract_paths_maybe_set_cwd(
                 .ignore_patterns
                 .as_ref()
                 .is_none_or(|patterns| !patterns.excludes_input_path(path, &cwd))
+        })
+        .filter(|path| {
+            // Only paths we expanded into roots ourselves are eligible for --ignore-dirs here,
+            // explicitly passed top-level paths are never ignored.
+            if !paths_were_expanded || walk_options.ignore_dirs.is_empty() {
+                return true;
+            }
+            !gix::path::realpath_opts(path, &cwd, 32)
+                .is_ok_and(|real| walk_options.ignore_dirs.contains(&real))
         })
         .collect())
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -445,13 +445,12 @@ fn extract_paths_maybe_set_cwd(
                 .is_none_or(|patterns| !patterns.excludes_input_path(path, &cwd))
         })
         .filter(|path| {
-            // Only paths we expanded into roots ourselves are eligible for --ignore-dirs here,
-            // explicitly passed top-level paths are never ignored.
             if !paths_were_expanded || walk_options.ignore_dirs.is_empty() {
                 return true;
             }
-            !gix::path::realpath_opts(path, &cwd, 32)
-                .is_ok_and(|real| walk_options.ignore_dirs.contains(&real))
+            let is_ignored = gix::path::realpath_opts(path, &cwd, 32)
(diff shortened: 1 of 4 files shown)


VXNCXNX:
The macOS journey test was failing on my `$PWD/dir` argument: the sandbox lives under a symlinked temp path there, so the absolute form did not line up with the canonicalized ignore list. Switched both cases to the relative `-i dir`, which exercises the same code path without depending on how the shell spells the working directory. Snapshots are unchanged.

VXNCXNX:
Correction to my last comment: the relative form was not the answer, since `canonicalize_ignore_dirs` runs before the chdir, so a relative `-i` resolves against the wrong directory. Now using `$(pwd -P)`, which resolves the symlink the same way `gix::path::realpath` does, so the macOS temp path lines up. Verified locally against a symlinked working directory, which reproduces the macOS condition.

VXNCXNX:
Third time, and this one is a real bug in the fix rather than in the test. On macOS `std::env::current_dir()` keeps the `/var` symlink while `canonicalize_ignore_dirs` resolves to `/private/var`, so passing that cwd into `realpath_opts` produced a path that could never match the ignore set. It now calls `gix::path::realpath` on the expanded root, which resolves the working directory itself. Reproduced locally by running from a symlinked directory, where the old code also failed to ignore.

VXNCXNX:
I should stop guessing at this in public, so here is the honest state.

The macOS journey test still fails, and I do not have a Mac to debug it on. Three attempts from the CI log alone were all wrong: the argument now reaches the binary as a correctly resolved `/private/var/...` path, and the directory is still measured, so something about how the expanded root is resolved differs there in a way I cannot see from here.

On Linux the fix works and the journey tests pass, including a run from a symlinked working directory, which is the closest I can get to the macOS shape. The unit-test route is not open: `extract_paths_maybe_set_cwd` changes the process directory, and a test doing that races `journeys_readonly::once_waits_for_replayed_refresh_to_finish` about one run in five.

Happy to drop the two journey cases so the suite is green and leave the source fix, or to leave it as is if you would rather see it fail until it is right. If you can paste what `dua -i $(pwd -P)/dir . --log-file /tmp/dua.log` logs on a Mac, that would probably settle it in one round.

VXNCXNX:
Heads up before you spend time on this: the macOS lane is failing and it is my change, not a flake.

```
$ dua aggregate -i /private/var/.../T/sandbox.XXXXXX.BI8YLUjCP3/dir .
 <SIZE> MB dir
 <SIZE> MB total
```

`dir` should not be there. Linux and Windows both pass, so the filter is not matching only where `/var` resolves through the `/private` symlink.

I cannot reproduce it. I built the equivalent locally, a real directory plus a symlinked parent, entered through the symlink, and passed `-i "$(pwd -P)/dir"`, and the ignore applies correctly on Linux both through the symlink and physically. So I do not have a verified diagnosis, only a suspicion: my filter calls `gix::path::realpath`, which takes its cwd from `std::env::current_dir()` internally, while the repo's own `ignore_directory` in `common.rs` calls `realpath_opts(path, cwd, 32)` with the cwd passed in explicitly. Reusing the existing helper instead of rolling a second resolution would be better regardless, but I would be guessing at whether that is the actual macOS difference.

I would rather say that than push a speculative fix and have you review something I have not run. If you have a moment to say whether that hunch is in the right area I will take it from there, and otherwise I am happy for this to sit until I can get on a Mac and check properly.

The rest of the change and the `--ignore-dirs is never applied to explicitly given paths` half of the behaviour are verified.

*Disclosure: written with AI assistance (Claude Code). The failure above is quoted from your CI, and the Linux non-reproduction is mine.*

Byron:
Thanks a lot.

PR #374, opened 2026-08-17
Skip path resolution without ignored directories

tamird:
Prepared macOS and Windows aggregation roots resolve every child path even when the ignored-directory list is empty. Path resolution checks for symbolic links and therefore adds one filesystem lookup per root.

Share the existing traversal predicate with both aggregation input paths. Its empty-list short circuit skips path resolution, and prepared roots use borrowed filenames without allocating paths.

This reduces duplication and fixes a tiny missed optimization from f9b6200c8.


Byron:
Thanks, good catch!

PR #375, opened 2026-08-17
Skip clone metadata for apparent sizes

tamird:
Combining apparent-size reporting with APFS clone deduplication still requests extended clone metadata even though logical sizes never use it. Explicit file roots also perform unnecessary attribute lookups.

Disable clone metadata whenever apparent sizes are requested while preserving both accepted command-line flags. Cover all four flag combinations in the shared traversal-options constructor.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -346,7 +346,7 @@ fn walk_options_from(traversal: &options::TraversalArgs) -> Result<dua::WalkOpti
         ignore_patterns: dua::IgnorePatterns::from_files(&traversal.ignore_from)?,
         metadata_options: dua::TraversalOptions {
             #[cfg(target_os = "macos")]
-            apfs_clone_metadata: traversal.deduplicate_apfs_clones,
+            apfs_clone_metadata: traversal.deduplicate_apfs_clones && !traversal.apparent_size,
         },
     };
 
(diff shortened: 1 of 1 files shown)


Byron:
Good catch.

It seems to not make much of a difference, but it is a difference.

```
❯ hyperfine -i './target/release/dua a ../../ --deduplicate-apfs-clones --apparent-size' './target/release/dua a ../../ --deduplicate-apfs-clones'
Benchmark 1: ./target/release/dua a ../../ --deduplicate-apfs-clones --apparent-size
  Time (mean ± σ):      6.074 s ±  0.117 s    [User: 1.621 s, System: 27.379 s]
  Range (min … max):    5.922 s …  6.293 s    10 runs

Benchmark 2: ./target/release/dua a ../../ --deduplicate-apfs-clones
  Time (mean ± σ):      6.281 s ±  0.148 s    [User: 1.612 s, System: 27.341 s]
  Range (min … max):    6.048 s …  6.560 s    10 runs

Summary
  ./target/release/dua a ../../ --deduplicate-apfs-clones --apparent-size ran
    1.03 ± 0.03 times faster than ./target/release/dua a ../../ --deduplicate-apfs-clones
```

PR #376, opened 2026-08-17
Avoid materiailizing vector

tamird:
This saves an allocation at the expense of delaying reporting of traversal errors.

This previously changed from streaming to eager in 0acd5fa42 during review of #371.

PR note: if eager error reporting is important I'd be happy to change this into a comment addition instead.


Byron:
Why not, thanks!

PR #377, opened 2026-08-17
Shrink native macOS traversal metadata

tamird:
Optional APFS fork metadata adds a separate discriminant to every native directory entry, including traversals with clone deduplication disabled. Store the effective data allocation and optional clone identifier directly instead.

This reduces native metadata from 88 to 80 bytes and directory entries from 144 to 136 bytes without changing the public accessors. Assert both layouts and preserve resource-fork fallback behavior.

PR note: I'm not sure this is worth it; will let you decide.


Byron:
Thanks a lot!

Issue #289, opened 2025-07-17
Option to pipe disk usage into flamegraph


theimpostor:
Flamegraphs are great for profiling but can be used for [disk space also](https://www.brendangregg.com/blog/2017-02-05/file-system-flame-graph.html). It would be nice if dua had a CLI option to output results in the folded stack format, so that it could be piped into [flamegraph.pl](https://github.com/brendangregg/FlameGraph/blob/master/flamegraph.pl) (or more modern alternatives like [inferno](https://github.com/jonhoo/inferno)).

The output format is pretty simple, every file / directory printed on a single line followed by a space and the size in bytes. The path separators are replaced with `;`, and whitespace and other special chars are replaced with `_`. The output does not need to be sorted.

It would also be nice if hard linked files were only counted the first time, but not necessary (I don't think the original [files.pl](https://github.com/brendangregg/FlameGraph/blob/master/files.pl) did that).

theimpostor:
This can probably be closed due to #378, the stacks subcommand works great, thank you!

PR #378, opened 2026-08-18
Add 'stacks' subcommand for folded/flame-graph output (#289)

ChrisJr404:
Adds a `stacks` subcommand that prints the whole traversal as folded stacks, so a `dua` run can be turned into a disk-usage flame graph without leaving the tool. This implements the request in #289.

The folded (a.k.a. "collapsed") format is what flame-graph tools like [inferno](https://github.com/jonhoo/inferno) and Brendan Gregg's `flamegraph.pl` read on stdin: one line per entry, its path with `;` between components, a space, and its size in bytes. That makes the usual pipe work:

```bash
dua stacks | inferno-flamegraph > disk-usage.svg
```

A couple of design notes:

- Rather than only emitting files (as Gregg's `files.pl` does), every entry gets a line carrying its *own* size: a directory contributes just the size of its directory entry, because the bytes it contains already show up on the lines of the things inside it. That avoids double-counting when the flame-graph tool sums a stack against its parents, and it means the folded lines sum to exactly the same total the other commands report — I added a test that checks this against a fresh traversal. Zero-sized entries are left out since they add nothing to a graph.
- It reuses the existing traversal, so all the traversal options carry over (`--apparent-size`, `--ignore-from`, `--count-hard-links`, threads, and so on). Hard links are counted once by default, as elsewhere.
- Entry names have the `;` frame separator and any control characters replaced with `_` so a single odd filename can't corrupt the line-based format.

I went with a subcommand rather than a flag on `aggregate` because the folded output needs the full in-memory tree, which `aggregate` doesn't build — but I'm happy to reshape the surface (name, flag vs. subcommand) if you'd prefer something else.

Example output:

```
$ dua stacks -A some_dir
a 80
a;big 5000
a;b 60
a;b;small 100
top 3
```

Testing: added four unit tests in `src/stacks.rs` covering the nested-file layout, the sum-equals-total invariant, a single-file input, and `;`-in-name sanitization. `cargo test` is green (77 tests), `cargo +nightly fmt --check` is clean, and `cargo +nightly clippy` produces no new warnings from this change. Also smoke-tested by hand that `dua stacks -A` and `dua aggregate -A --format bytes` report the same total, and that the output pipes cleanly.


Byron:
Can you please upload the SVG to this PR? I want to take a look. Please beware of PII. Thanks so much.

Byron:
Actually it's a fantastic feature!

```
cargo run --release -- a --stack ../.. | inferno-flamegraph > flame.svg
```
flame

```
cargo run --release -- a --depth 2 --stack ../.. | inferno-flamegraph > flame.svg
```
flame

PR #379, opened 2026-08-21
Match macOS `du` allocation accounting

nshcr:
> [!NOTE]
> **AI development disclosure**
> 
> Both the code changes in this PR and this description were produced by Codex under my direction. I defined the scope and guided the investigation; Codex performed the data collection, cross-version comparisons, source and history tracing, implementation, test development and execution, validation, and final drafting.
> 
> I manually reviewed the relevant primary sources and found no serious issues. If this approach is incorrect or does not fit `dua`'s direction, please feel free to modify or reject it, or let me know how it should be improved!

## Background

Hi! I found this issue unexpectedly while using `dua` on macOS 26. When scanning an OrbStack directory mounted through NFS, `dua`'s default allocated-size total did not match the equivalent result from `/usr/bin/du`.

At first, this looked like a single NFS regression. Testing several `dua` releases showed that the discrepancy had actually changed direction:

- v2.41.1 and earlier overcounted the tree;
- v2.42.0 and v2.42.1 undercounted it.

Further checks ruled out hard-link deduplication, traversal concurrency, and cancellation between subtrees. The discrepancy came from the allocated-size metadata returned by three different macOS query contracts.

This PR assumes that matching the system `du` behavior may be desirable for `dua`, at least for default allocated-size accounting on macOS. It therefore follows the metadata contract used by Apple FTS - the traversal implementation behind macOS `/usr/bin/du` - rather than adding an OrbStack/NFS-specific workaround.

## Reproduction

The comparison uses equivalent default allocated-size modes:

```bash
dua aggregate -f bytes "$NFS_ROOT"
BLOCKSIZE=512 /usr/bin/du -s "$NFS_ROOT"
```

`dua` reports bytes directly. With `BLOCKSIZE=512`, macOS `du` reports 512-byte blocks:

```text
/usr/bin/du: 12,442,673 × 512 = 6,370,648,576 bytes
```

The main comparison was run sequentially against the same mounted tree:

| Implementation      | Reported bytes | Delta vs `/usr/bin/du` |
| ------------------- | -------------: | ---------------------: |
| v2.41.0             |  6,628,753,408 | +258,104,832 (+4.051%) |
| v2.41.1             |  6,628,753,408 | +258,104,832 (+4.051%) |
| v2.42.0             |  6,357,613,568 |  -13,035,008 (-0.205%) |
| v2.42.1             |  6,357,612,032 |  -13,036,544 (-0.205%) |
| Initial fixed build |  6,370,648,576 |                      0 |
| `/usr/bin/du`       |  6,370,648,576 |              Reference |

These runs traversed 140,829 entries and reported the same 13 inaccessible paths. A later v2.39.1 check reproduced the v2.41.x byte total, although that older release propagated traversal errors differently. This also shows that the v2.40 walker was not the origin of the overcount.

OrbStack is a live filesystem, so absolute totals describe individual snapshots. The relevant evidence is the stable version-to-version difference under matching entry and error counts, followed by repeated exact agreement between the fix and `du` on later snapshots.

## Root cause

Before v2.42.0, `dua` ultimately accounted allocated size using per-entry Unix metadata and `st_blocks × 512`. That is a conventional Unix definition, but it does not guarantee the same filesystem query contract as Apple FTS. On this NFS mount, the per-path metadata query returned a total 258,104,832 bytes larger than `/usr/bin/du`.

PR #367 introduced a native macOS walker using `getattrlistbulk`. Its original implementation:

- requested a reduced common/file/directory attribute set;
- called `getattrlistbulk(..., options = 0)`;
- trusted records without first proving that the returned attributes were sufficient to synthesize a complete `stat`;
- used directory allocation from bulk records, whereas Apple FTS always performs a path lookup for directories.

These differences were not exposed by the existing local-filesystem tests, but on this NFS mount they changed the discrepancy from an overcount to an undercount of 13,035,008 bytes.

PR #369 then reused the same bulk metadata for aggregation roots. This was not the primary cause, but it extended the affected path and reduced the total by another 1,536 bytes on this snapshot.

## Implementation

This PR makes `dua`'s macOS bulk walker follow the relevant Apple FTS contract:

- Request the complete common and file attribute sets needed to synthesize `stat`.
- Leave directory attributes unset, because FTS does not synthesize directory `stat` fields from bulk records.
- Use `FSOPT_PACK_INVAL_ATTRS` for full metadata enumeration.
- Decode packed-invalid fields in request order while separately retaining the returned-attribute bitmap.
- Accept bulk metadata only for FTS-recognized vnode types when the required attributes were actually returned.
- Fall back to no-follow path metadata for directories, firmlinks, unknown vnode types, and incomplete records.
- Round allocation bytes to 512-byte accounting units, matching how FTS constructs `st_blocks`.
- Preserve the existing list-only permission retry and ordinary directory-reading fallback.
- Keep the bulk fast path for complete non-directory records.

The branch is based on #371. Its optional APFS data-fork allocation, clone ID, extended flags, and `FSOPT_ATTR_CMN_EXTENDED` behavior are retained as extensions of the same packed layout. Missing APFS extension attributes disable clone metadata without invalidating an otherwise complete FTS-compatible record.

The implementation does not inspect the filesystem name and contains no NFS-specific branch.

## Validation

The following checks passed on the current branch:

```text
cargo fmt --all -- --check
cargo clippy --workspace --all-targets --all-features -- -D warnings
cargo test --workspace --all-targets --all-features
git diff --check
```

All 126 automated tests passed. The macOS coverage includes bulk metadata, APFS clones and data forks, resource forks, symlinks, hard links, permissions, buffer refills, fallback behavior, FTS/APFS request shapes, vnode validation, returned-attribute completeness, and allocation rounding.

Real OrbStack NFS validation confirmed that:

- the default total matched `/usr/bin/du` exactly;
- the `ubuntu` and `docker` subtrees matched independently, ruling out cancellation between positive and negative errors;
- `--count-hard-links` matched `/usr/bin/du -l` exactly;
- 1, 4, and 8 threads produced identical totals, entry counts, and errors;
- `--apparent-size` remained unchanged from v2.42.1;
- the same 13 permission errors were preserved.

After rebasing onto #371, the current head was checked again on a later filesystem snapshot:

```text
Default accounting:
dua                 6,370,499,072 bytes
/usr/bin/du         6,370,499,072 bytes

Counting hard links:
dua                 7,586,685,440 bytes
/usr/bin/du -l      7,586,685,440 bytes
```

The absolute totals changed with the live filesystem, but equality with the corresponding `du` mode remained exact. The rebased head also produced the same result with one thread and the default eight threads.

No automated real-NFS fixture is included. Unit tests protect the query, parsing, validation, fallback, and rounding rules; the OrbStack runs provide end-to-end evidence for the original issue.

## Performance

A controlled benchmark performed before the rebase onto #371 compared v2.42.1 with the initial fix. Both binaries used Rust 1.95.0 and the same release/all-features configuration. After one warm-up per binary, each workload ran 12 times in alternating order:

| Workload | v2.42.1 mean ± sd | Fix mean ± sd | Median, before → after |
|---|---:|---:|---:|
| OrbStack NFS, ~140k entries and 13 errors | 0.387 ± 0.014 s | 0.403 ± 0.013 s | 0.380 → 0.400 s |
| APFS fixture with 100,101 directories | 1.790 ± 0.138 s | 1.725 ± 0.162 s | 1.820 → 1.755 s |

The NFS workload was approximately 17 ms slower on average. The APFS difference was smaller than the observed run-to-run variation. These two workloads did not show a material regression, but they do not cover every filesystem, tree shape, or cache state.

This benchmark predates the rebase onto #371. The rebased head has passed correctness and APFS clone tests, but it has not been re-benchmarked against the current `main`.

## Compatibility and limitations

- The code changes are limited to macOS; Linux and Windows paths are unchanged.
- The accounting contract applies to all macOS filesystems using the bulk path, not only NFS.
- Complete non-directory records retain the bulk fast path.
- Directories now use path metadata, matching Apple FTS; this may have different costs on directory-heavy workloads.
- No CLI, public API, or serialization changes are introduced relative to the current base.
- This PR does not address NFS server availability, retries, timeouts, or mount recovery.
- It does not change `--apparent-size` or attempt byte-for-byte parity with macOS `du -A`.
- The real-world validation covers OrbStack NFS and should not be treated as certification for every NFS server or mount configuration.

## References

- [PR #35: introduce real/apparent size accounting](https://github.com/Byron/dua-cli/pull/35) and the [v2.41.1 Unix allocated-size call site](https://github.com/Byron/dua-cli/blob/90a59e15a681d50a64c0f86483100b257297f5a7/src/traverse.rs#L393-L402)
- [PR #367: introduce the native macOS bulk walker](https://github.com/Byron/dua-cli/pull/367), including its [attribute request](https://github.com/Byron/dua-cli/blob/7d115c014eff2f85ab60e82266f4f1beb4782598/crates/dua-lib/src/macos/attributes.rs#L22-L42) and [`getattrlistbulk` call](https://github.com/Byron/dua-cli/blob/7d115c014eff2f85ab60e82266f4f1beb4782598/crates/dua-lib/src/macos.rs#L211-L219)
- [PR #369: reuse bulk metadata for aggregation roots](https://github.com/Byron/dua-cli/pull/369)
- [PR #371: add optional APFS clone accounting](https://github.com/Byron/dua-cli/pull/371)
- [Apple FTS: attribute request and bulk enumeration](https://github.com/apple-oss-distributions/Libc/blob/71bbe350ab79eef58113991d817ccc6165061a64/gen/fts.c#L778-L824)
- [Apple FTS: vnode validation, fallback behavior, and `st_blocks` construction](https://github.com/apple-oss-distributions/Libc/blob/71bbe350ab79eef58113991d817ccc6165061a64/gen/fts.c#L903-L994)
- [Apple `du`: FTS traversal and `st_blocks` aggregation](https://github.com/apple-oss-distributions/file_cmds/blob/659a8a301e2acf0343f8b8673a154a2ca4d07084/du/du.c#L330-L412)
- [Apple `getattrlist(2)` documentation](https://github.com/apple-oss-distributions/xnu/blob/main/bsd/man/man2/getattrlist.2)

### DIFF
--- a/crates/dua-lib/src/macos/attributes.rs
+++ b/crates/dua-lib/src/macos/attributes.rs
@@ -357,6 +357,10 @@ pub(super) fn parse_record(
     Ok(record)
 }
 
+/// Return whether a requested attribute occupies space in the record.
+///
+/// Packed-invalid records contain default values even when the returned-attribute mask marks the
+/// attribute unavailable; ordinary records omit unavailable attributes entirely.
 fn attribute_is_packed(
     returned: libc::attrgroup_t,
     attribute: libc::attrgroup_t,
--- a/crates/dua-lib/src/macos/attributes.rs
+++ b/crates/dua-lib/src/macos/attributes.rs
@@ -12,7 +12,11 @@ use std::{
 pub(super) const VNON: u32 = 0;
 pub(super) const VREG: u32 = 1;
 pub(super) const VDIR: u32 = 2;
+pub(super) const VBLK: u32 = 3;
+pub(super) const VCHR: u32 = 4;
 pub(super) const VLNK: u32 = 5;
+pub(super) const VSOCK: u32 = 6;
+pub(super) const VFIFO: u32 = 7;
 // These `<sys/stat.h>` filesystem flags and accounting units are absent from `libc`.
 pub(super) const SF_FIRMLINK: u32 = 0x0080_0000;
 pub(super) const STAT_BLOCK_BYTES: u64 = 512;
@@ -22,14 +26,19 @@ const EF_MAY_SHARE_BLOCKS: u64 = 0x0000_0001;
 const ATTR_CMN_ERROR: libc::attrgroup_t = 0x2000_0000;
 const NANOS_PER_SECOND: u32 = 1_000_000_000;
 
-const COMMON_ATTRIBUTES: libc::attrgroup_t = libc::ATTR_CMN_RETURNED_ATTRS
+const FTS_COMMON_ATTRIBUTES: libc::attrgroup_t = libc::ATTR_CMN_RETURNED_ATTRS
     | libc::ATTR_CMN_NAME
     | libc::ATTR_CMN_DEVID
     | libc::ATTR_CMN_OBJTYPE
+    | libc::ATTR_CMN_CRTIME
     | libc::ATTR_CMN_MODTIME
+    | libc::ATTR_CMN_CHGTIME
+    | libc::ATTR_CMN_ACCTIME
+    | libc::ATTR_CMN_OWNERID
+    | libc::ATTR_CMN_GRPID
+    | libc::ATTR_CMN_ACCESSMASK
     | libc::ATTR_CMN_FLAGS
-    | libc::ATTR_CMN_FILEID
-    | ATTR_CMN_ERROR;
+    | libc::ATTR_CMN_FILEID;
 // XNU's `vfs_attrlist.c` `LIST_DIR_ATTRS` authorizes this subset with LIST_DIRECTORY alone.
 // Metadata attributes additionally require SEARCH, which a readable directory need not grant.
 const LIST_ONLY_ATTRIBUTES: libc::attrgroup_t = libc::ATTR_CMN_RETURNED_ATTRS
@@ -37,12 +46,11 @@ const LIST_ONLY_ATTRIBUTES: libc::attrgroup_t = libc::ATTR_CMN_RETURNED_ATTRS
     | libc::ATTR_CMN_OBJTYPE
     | libc::ATTR_CMN_FILEID
     | ATTR_CMN_ERROR;
-const DIRECTORY_ATTRIBUTES: libc::attrgroup_t = libc::ATTR_DIR_LINKCOUNT
-    | libc::ATTR_DIR_MOUNTSTATUS
-    | libc::ATTR_DIR_ALLOCSIZE
-    | libc::ATTR_DIR_DATALENGTH;
-const FILE_ATTRIBUTES: libc::attrgroup_t =
-    libc::ATTR_FILE_LINKCOUNT | libc::ATTR_FILE_ALLOCSIZE | libc::ATTR_FILE_DATALENGTH;
+const FTS_FILE_ATTRIBUTES: libc::attrgroup_t = libc::ATTR_FILE_LINKCOUNT
+    | libc::ATTR_FILE_ALLOCSIZE
+    | libc::ATTR_FILE_IOBLOCKSIZE
+    | libc::ATTR_FILE_DEVTYPE
+    | libc::ATTR_FILE_DATALENGTH;
 const APFS_FILE_ATTRIBUTES: libc::attrgroup_t = libc::ATTR_FILE_DATAALLOCSIZE;
 const APFS_EXTENDED_ATTRIBUTES: libc::attrgroup_t =
     libc::ATTR_CMNEXT_CLONEID | libc::ATTR_CMNEXT_EXT_FLAGS;
@@ -96,11 +104,11 @@ pub(super) fn requested_attributes(list_only: bool, include_apfs: bool) -> libc:
         commonattr: if list_only {
             LIST_ONLY_ATTRIBUTES
         } else {
-            COMMON_ATTRIBUTES
+            FTS_COMMON_ATTRIBUTES
         },
         volattr: 0,
-        dirattr: if list_only { 0 } else { DIRECTORY_ATTRIBUTES },
-        fileattr: if list_only { 0 } else { FILE_ATTRIBUTES },
+        dirattr: 0,
+        fileattr: if list_only { 0 } else { FTS_FILE_ATTRIBUTES },
         forkattr: 0,
     };
     if include_apfs && !list_only {
@@ -124,6 +132,8 @@ pub(super) fn root_clone_attributes() -> libc::attrlist {
 
 #[derive(Default)]
 pub(super) struct ParsedRecord {
+    returned_common: libc::attrgroup_t,
+    returned_file: libc::attrgroup_t,
     pub(super) file_name: Option<OsString>,
     pub(super) device: Option<u64>,
     pub(super) object_type: Option<u32>,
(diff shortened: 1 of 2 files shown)


Byron:
Hi there, nice to see you here!

In my unscientific tests, and with my dataset which contained orbstack, I see about a 2% reduction in performance:

```
❯ hyperfine -i './target/release/dua -f bytes ~/Library/' 'dua -f bytes ~/Library/'
Benchmark 1: ./target/release/dua -f bytes ~/Library/
  Time (mean ± σ):      2.961 s ±  0.135 s    [User: 0.564 s, System: 9.187 s]
  Range (min … max):    2.795 s …  3.222 s    10 runs

  Warning: Ignoring non-zero exit code.

Benchmark 2: dua -f bytes ~/Library/
  Time (mean ± σ):      2.898 s ±  0.169 s    [User: 0.522 s, System: 8.803 s]
  Range (min … max):    2.760 s …  3.356 s    10 runs

  Warning: Ignoring non-zero exit code.

Summary
  dua -f bytes ~/Library/ ran
    1.02 ± 0.08 times faster than ./target/release/dua -f bytes ~/Library/
```

```
❯ hyperfine -i './target/release/dua -f bytes ~/dev' 'dua -f bytes ~/dev'
Benchmark 1: ./target/release/dua -f bytes ~/dev
  Time (mean ± σ):      6.349 s ±  0.218 s    [User: 1.785 s, System: 28.204 s]
  Range (min … max):    6.132 s …  6.801 s    10 runs

Benchmark 2: dua -f bytes ~/dev
  Time (mean ± σ):      6.169 s ±  0.428 s    [User: 1.647 s, System: 27.224 s]
  Range (min … max):    5.921 s …  7.331 s    10 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Summary
  dua -f bytes ~/dev ran
    1.03 ± 0.08 times faster than ./target/release/dua -f bytes ~/dev
```

In the name of correctness, knowing that `du` probably knows better, I think this should be merged nonetheless.

Byron:
Thanks again, it's much appreciated!

Now that I reviewed it, and looking at the changes to the previous version, I realize how much I am out of my depth so I am mostly just… reviewing it very generically, without much I can truly add.

As a follow-up, it would be interesting to learn if the performance reduction can be countered somehow - my feeling is that the bulk calls are now a little bit slower, but I didn't profile a before-after knowing that it won't be trivial to find out. Also I didn't let an agent loose on the task, in fear it might go off the rails hunting a phantom 😅.

nshcr:
Thanks for merging this PR, and for reviewing and benchmarking it!

The fix reimplements parts of Apple FTS / `du`'s metadata and accounting contract in Rust. GPT-5.6 Sol handled the source comparison and integration into `dua`. I mainly verified the added attributes and constants against Apple's sources and reviewed the overall logic as far as I could - being fluent in both Rust and C is a little beyond my superpowers :p. It also went through an additional Codex review before submission.

I hope the report made clear that the NFS discrepancy was real and that the fix resolved it in the tested environment. I happened to notice it around the v2.42.0 release: the versions immediately before and after it reported strangely different sizes for the same OrbStack tree, and neither matched `du`.

So I see this PR as a report with a fix attached. Matching `du` matters for a disk-usage tool; even a small unexplained difference can make users question the results.

The 2-3% slowdown in your benchmarks is understandable since the corrected contract does more metadata work. Given the much larger performance gain from v2.42.0, it seems like a reasonable correctness trade-off, although improving it further is beyond my current expertise too.

Byron:
Thanks, you are welcome, it's the least I can do.

> So I see this PR as a report with a fix attached. Matching `du` matters for a disk-usage tool; even a small unexplained difference can make users question the results.

While I like matching `du`, in theory, at least in theory, it doesn't matter to `dua` as it's a tool to help find 'big' files, and delete them. For that it doesn't matter if these are reported as a little larger than they really are.
However, I think the makeover leads to macOS attributes code that looks more mature, and given how little I understand of it, it seems like the more secure choice moving forward, and it's worth to pay the price for it.

After all, it's nothing one will notice nor will it materially affect anything.

Now, the timer is reset and I wonder how many days it will be till the next release  😅 - everyone, give your agents some rest please 😁.

PR #380, opened 2026-08-25
Add aggregate --depth for indented tree output

### LINKED ISSUE #264: Non-interactive tree output
By default `dua some_dir` reports a flat file hierarchy. However, it would be useful to generate a tree, but non-interactively. My use case is sharing a disk usage report with someone, which can't be done in interactive mode, and the current behaviour isn't informative enough.

I wonder if a `--depth <int>` flag could be added to the uninteractive command that allows a tree output? It should default to 1 which is the current behaviour.

ChrisJr404:
Closes #264.

This adds a `--depth N` option to the `aggregate` subcommand that prints an indented tree instead of the flat listing, which is handy for sharing a nested disk usage report without dropping into interactive mode. The inputs are the first level, so `--depth 1` lists exactly what the flat output does today, and higher values walk further down.

It reuses the existing traversal tree, sorts each level the same way the flat listing does (and respects `--no-sort` and `--no-total`), and colors directories like the rest of the output. Docs and tests are included.

I put the flag on `aggregate` alongside `--no-total` and `--no-sort` rather than at the root, but happy to expose it at the top level too if you'd rather.


Byron:
Thanks, I like it.

Note to self: Should merge only after https://github.com/Byron/dua-cli/pull/373 as that will use colors only in a terminal.

PR #383, opened 2026-08-29
Stream folded stacks during traversal

### LINKED ISSUE #382: `aggregate --stack` buffers the whole tree, so memory scales with entry count on very large filesystems
### Summary

`dua aggregate --stack` holds the complete traversal tree in memory and writes nothing until the
walk finishes. On a filesystem with tens of millions of entries this reaches several gigabytes of
resident memory, which is enough to be killed by a container memory limit that is otherwise
generous.

The folded-stack output itself is a stream of independent lines. That suggests the tree does not
have to be complete before the first line can be written.

### Environment

- `dua` 2.43.0, installed with `cargo install --locked --no-default-features dua-cli@2.43.0`
- Linux x86_64, inside a container
- The target is a large shared network-backed volume, mounted locally

### What I observed

The command under test:

```sh
dua aggregate -x --stack /path/to/mount
```

The tree:

| Property | Value |
|---|---|
| Entries reported | about 20.5 million |
| Total size | about 2.8 TB |
| Folded output lines | about 20.5 million |
| Walk duration | about 2 minutes 30 seconds |

Two results:

1. In a container limited to 2 GiB, the process was OOM-killed part way through the walk. It
   produced no output at all, so a consumer downstream of the pipe saw an empty stream.
2. In a container with a large limit, the same command completed. Peak memory usage for the
   control group holding only this process reached about 4.2 GiB.

No output appeared until the walk was essentially complete. A consumer reading the pipe stayed idle
for the whole traversal and then received all lines at once.

### How I measured

I sampled the control group's memory accounting once per second and kept the maximum. That figure
includes reclaimable page cache, so it overstates the anonymous memory the process actually needs.
I am reporting it as an upper bound rather than as an exact heap figure.

The OOM kill at 2 GiB is the firmer signal. Page cache is reclaimable under pressure, so a kill at
that limit indicates the anonymous allocation alone exceeded 2 GiB well before the walk finished.

### Why I think this is the tree, not the output

Reading `src/stacks.rs` in 2.43.0, `stacks()` drives the traversal to completion, integrating every
event into a `Traversal`, and only then calls `write_stacks()` to walk the finished tree and emit
folded lines. Nothing is written while the traversal is running.

Each folded line is self-contained: a path with `;` between components, a space, and that entry's
own size in bytes. Ordering does not appear to matter to the flame-graph tools that consume this
format.

### What would help

Some way to keep memory bounded by traversal depth rather than by total entry count. A few shapes
that would each solve it, in rough order of preference:

1. Emit folded lines during the traversal instead of after it. A directory's own size is known once
   its children have been accounted for, so a depth-first walk could write and then release each
   completed subtree.
2. An opt-in streaming mode, if changing the default behaviour of `--stack` is unwelcome.
3. Documentation of the expected memory cost per entry, so operators can size a job up front.

I have not measured whether `--depth` bounds memory during the walk or only prunes the finished
tree. If it does bound the walk, saying so in `--help` would already help people who hit this.

### Impact

This is not blocking me. I worked around it by giving the job a much larger memory limit. I am
filing it because the failure mode is unhelpful: the process dies part way through a long walk,
emits nothing, and the pipeline downstream cannot tell an empty filesystem from a killed producer.

Thanks for `dua`, and for adding `--stack`. It replaced a local patch I had been carrying.

theimpostor:
Closes #382.

## Summary

- consume completion-order traversal events directly instead of retaining a `Traversal` graph
- write each unbounded folded stack before requesting the next event
- preserve hard-link filtering, filesystem boundaries, ignore handling, platform-specific allocation accounting, progress, and error counts
- keep `--depth` output compact by aggregating only entries collapsed onto the depth boundary
- add a regression test proving output happens before the next traversal event is consumed

## Verification

- `make check unit-tests journey-tests check-fmt clippy`
- compared sorted old/new folded output on a nested fixture, both unlimited and `--depth 2`; output was identical

On a release build scanning 200,000 empty local files with `-A -t 1`:

| | First output | Peak RSS | Total |
|---|---:|---:|---:|
| `main` | 581 ms | 101,760 KiB | 591 ms |
| this PR | 13 ms | 22,600 KiB | 784 ms |

This cached, single-worker, empty-file case exposes the completion walker's small-batch scheduling overhead. I also tested parent-first event consumption: it reduced total time to about 0.72 s but retained the same ~101,760 KiB peak because the wide directory remained buffered as one batch, so this uses completion order to achieve the issue's memory goal. On I/O-bound storage the scheduling overhead should be a smaller fraction of the walk.

### DIFF
--- a/src/stacks.rs
+++ b/src/stacks.rs
@@ -1,19 +1,36 @@
 use crate::aggregate::TraversalProgress;
-use crate::traverse::{BackgroundTraversal, EntryData, Traversal, Tree, TreeIndex};
-use crate::{WalkOptions, WalkResult};
+use crate::{InodeFilter, WalkOptions, WalkResult, WalkRoot, crossdev};
 use anyhow::Result;
 use bstr::ByteSlice;
-use petgraph::Direction;
+#[cfg(not(any(windows, target_os = "macos")))]
+use filesize::PathExt;
+use std::collections::HashMap;
+use std::ffi::OsStr;
 use std::io;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
-/// Traverse `paths` and write the tree to `out` as folded stacks, one entry per line, ready
-/// to pipe into flame-graph tools like [`inferno`](https://github.com/jonhoo/inferno).
+#[cfg(not(any(windows, target_os = "macos")))]
+fn size_on_disk(entry: &crate::walk::Entry, metadata: &crate::walk::Metadata) -> io::Result<u64> {
+    entry.path().size_on_disk_fast(metadata)
+}
+
+#[cfg(windows)]
+#[allow(clippy::unnecessary_wraps)]
+fn size_on_disk(entry: &crate::walk::Entry, metadata: &crate::walk::Metadata) -> io::Result<u64> {
+    Ok(if entry.file_type.is_dir() {
+        0
+    } else {
+        metadata.allocated_size()
+    })
+}
+
+/// Traverse `paths` and write entries to `out` as folded stacks as soon as they are discovered,
+/// ready to pipe into flame-graph tools like [`inferno`](https://github.com/jonhoo/inferno).
 ///
 /// Each line is an entry's path from the traversal root, with its components separated by `;`,
-/// followed by a single space and the entry's own size in bytes. A directory contributes only the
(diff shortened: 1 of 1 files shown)


Byron:
Superseded by https://github.com/Byron/dua-cli/pull/384.
Please take a look to see if this works for you just as well.

PR #384, opened 2026-08-29
Stream folded stacks during traversal

### LINKED ISSUE #382: `aggregate --stack` buffers the whole tree, so memory scales with entry count on very large filesystems

Byron:
## Tasks

This section is for Byron only. Models continuing this PR must not add, remove, check, uncheck, rename, or reorder checkboxes here.

- [ ] refackiew

----

Everything below this line was generated by `Codex GPT-5`.

Created by Codex on behalf of Byron. Byron will review before this is ready to merge.

Fixes #382

## Summary

- Stream unlimited-depth `aggregate --stack` lines as traversal events are integrated.
- Retain only root accounting instead of one graph node per filesystem entry.
- Preserve depth-limited aggregation, hard-link accounting, ignore behavior, and folded-path encoding.

## Validation

- `make check-pre-push`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- Regression test: output begins before a deep traversal finishes.
- Codex review of `deec7cc7`: no findings.

## Commit

- `deec7cc7 fix: stream folded stacks during traversal (#382)`

### DIFF
--- a/src/stacks.rs
+++ b/src/stacks.rs
@@ -1,9 +1,12 @@
 use crate::aggregate::TraversalProgress;
-use crate::traverse::{BackgroundTraversal, EntryData, Traversal, Tree, TreeIndex};
+use crate::traverse::{
+    BackgroundTraversal, EntryData, Traversal, TraversalEntry, TraversalEvent, Tree, TreeIndex,
+};
 use crate::{WalkOptions, WalkResult};
 use anyhow::Result;
 use bstr::ByteSlice;
 use petgraph::Direction;
+use std::ffi::OsStr;
 use std::io;
 use std::path::PathBuf;
 
@@ -14,6 +17,12 @@ use std::path::PathBuf;
 /// followed by a single space and the entry's own size in bytes. A directory contributes only the
 /// size of its own directory entry, as the sizes of everything it contains appear on the lines of
 /// the contained entries.
+///
+/// Without `max_depth`, each entry's own size is known when it is traversed, so its line can be
+/// written immediately without retaining the complete tree. With `max_depth`, sizes below the
+/// limit must be folded into their nearest visible ancestor. Because an ancestor is visited before
+/// its descendants, its final size is unknown until traversal finishes; the requested levels are
+/// therefore retained and written afterward instead of streamed.
 pub fn stacks(
     mut out: impl io::Write,
     err: Option<impl io::Write>,
@@ -22,6 +31,7 @@ pub fn stacks(
     max_depth: Option<usize>,
 ) -> Result<WalkResult> {
     let mut traversal = Traversal::new();
+    let stream = max_depth.is_none();
     // Mirror the interactive traversal so root nodes carry their input path as name.
     let pattern_roots = walk_options.ignore_patterns.as_ref().map(|_| paths.clone());
     let mut background = BackgroundTraversal::start(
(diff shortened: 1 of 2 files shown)


Byron:
Screenshot 2026-08-29 at 10 01 46
Screenshot 2026-08-29 at 10 01 52

While the graphs are exactly the same, the coloring isn't. That's probably not a concern here.

PR #404, opened 2026-09-22
docs: correct platform defaults in threads help

Likio3000:
The `--threads` help says its default is always 0 (logical processor count), while `DEFAULT_THREADS` is 8 on macOS and 0 elsewhere. Correct the help to distinguish these defaults and retain the meanings of explicit values 0 and 1.

Checked against the platform-specific constants and the clap default on the same field. This changes only a Rust documentation comment; no thread behavior changes. No build or project tests were run. Prepared and reviewed with AI assistance.


Byron (approved):
Thanks a lot, good catch!

